PR #3, opened 2016-05-06
Add `updated` date and time to each update in the updateinfo.xml.

goosemania:
The absence of the `updated` field may cause errors during sync of the repo
or during upload of the erratum.

### DIFF
--- a/rpm/assets/updateinfo.xml
+++ b/rpm/assets/updateinfo.xml
@@ -5,6 +5,7 @@
   <title>Sea_Erratum</title>
   <release>1</release>
   <issued date="2012-01-27 16:08:06"/>
+  <updated date="2012-01-27 16:08:06"/>
   <description>Sea_Erratum</description>
   <pkglist>
     <collection short="">
@@ -27,6 +28,7 @@
   <title>Bird_Erratum</title>
   <release>1</release>
   <issued date="2013-01-27 16:08:08"/>
+  <updated date="2013-02-27 17:00:00"/>
   <description>ParthaBird_Erratum</description>
   <pkglist>
     <collection short="">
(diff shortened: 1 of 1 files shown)


elyezer:
@goosemania seems that we have a really good test case here, should not Pulp report when that information is missing? That said I think we can add a counterpart test on Pulp Smash to ensure that Pulp identifies the missing of this information.

I agree adding this change, but we should make sure that Pulp provides useful informaiton if the information is not available.

Ichimonji10:
I glanced over the Pulp 2.9 test results this morning, and I got the impression that Pulp 2.9 does complain about this. Here's a section from a task report, in `['progress_report']['yum_importer']['errata']['error']`:

> time data '' does not match format '%Y-%m-%d %H:%M:%S', Unknown format: unable to convert "" to the datetime object, Fail to update the existing erratum <del>...</del>: Unable to parse the `updated` field

goosemania:
@elyezer, AFAIK, we can't be sure that every single erratum (or update in updateinfo.xml) will contain the `updated` field and that it will be in a proper format. But I think that we still want to be able to handle such erratum during the first sync or upload. Let's discuss it with the team.
Pulp will identify missing information during the update of the erratum. And I wrote [an issue](https://github.com/PulpQE/pulp-smash/issues/237) to test it.

@Ichimonji10, yes

elyezer:
@goosemania thanks for the explanation. Yes we can add the information but we should have a test ensuring that Pulp will handle the information missing properly. The issue you created will take care of it.

I'm OK on merging this.

Ichimonji10:
ACK

Thanks for the info, @goosemania.

Ichimonji10:
I can verify that `make rpm-fixtures` successfully executes on a Fedora 23 system with the commit in this PR applied.

elyezer:
:+1:

PR #4, opened 2016-05-09
Add ability to generate repository with updated or invalid updateinfo…

goosemania:
….xml.

PR #5, opened 2016-05-10
Add more make targets for RPM fixtures

Ichimonji10:
Add two new make targets:

* `rpm-fixtures-invalid-updateinfo` generates RPM fixtures with an
  invalid `updateinfo.xml` file.
* `rpm-fixtures-updated-updateinfo` generates RPM fixtures with an
  updated `updateinfo.xml` file.

These new fixture generation targets make use of the existing RPM
assets, meaning that no duplicate data sets need to be maintained.

This commit derived from https://github.com/PulpQE/pulp-fixtures/pull/4.
(Thank you, @goosemania.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,1 @@
-/*-fixtures/
+/*-fixtures*/
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,7 @@
-ALL = docker-fixtures rpm-fixtures
+ALL = docker-fixtures \
+    rpm-fixtures \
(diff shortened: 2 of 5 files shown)


Ichimonji10:
@goosemania Does this accomplish the same as #4? As far as I can tell, I think it does. I considered commenting on #4, but decided that simply showing what I have in mind may be more efficient.

elyezer:
Looks good to me

goosemania:
@Ichimonji10, I believe it does. Thank you!
Looks very good to me)

PR #7, opened 2016-05-11
Write all fixtures to a single directory

Ichimonji10:
Add a top-level directory named `fixtures`, and update the make file so
that all compiled files are placed in this directory. **This change is
breaking.** If a client used this command before this commit:

    make docker-fixtures

Then the client should use this command after this commit:

    make fixtures/docker

The former command created a directory named `docker-fixtures`, and the
latter command creates a directory at `fixtures/docker`.

The `all` target has been kept for backward compatibility, but it issues
a warning when successfully executed.

Fix https://github.com/PulpQE/pulp-fixtures/issues/6

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,1 @@
-/*-fixtures*/
+/fixtures/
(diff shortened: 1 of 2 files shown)


elyezer:
LGTM, I will make sure that the job also uses the new method.

elyezer:
There we go https://github.com/pulp/pulp_packaging/pull/156.

PR #8, opened 2016-05-17
- added dummy rpms

midnightercz:
dummy rpms for PR253 in pulp_smash


Ichimonji10:
PulpQE/pulp-smash#253 has been updated so that these files are not necessary. Closing.

If this issue is later re-opened, make sure to check the various metadata files and patches for sanity. Files like `comps.xml` may need to be updated.

PR #13, opened 2016-06-08
Document dependencies

Ichimonji10:
Each of the fixture generation scripts may depend on certain system
utilities being available and usable. Unfortunately, these dependencies
are not documented. Fix this omission by adding a "Dependencies" section
to the readme.

This new section makes use of a definition list. The markdown syntax
does not support definition lists. Convert the readme from markdown
syntax to reStructuredText syntax.

Fix: https://github.com/PulpQE/pulp-fixtures/issues/1

Fix: https://github.com/PulpQE/pulp-fixtures/issues/2


elyezer:
This is awesome, thanks for updating the documentation.

PR #14, opened 2016-06-17
Add new make target: fixtures/rpm-erratum

Ichimonji10:
The new make target generates a JSON erratum file that can be imported
into a Pulp RPM repository.


Ichimonji10:
Required by https://github.com/PulpQE/pulp-smash/pull/310

elyezer:
Thank you

rpm/gen-erratum.sh
```diff
@@ -0,0 +1,215 @@
...
+
+# Given a path to an RPM package, echo the package epoch.
+#
+# e.g. assets/walrus-5.21-1.noarch.rpm → 0
+get_rpm_epoch() {
+    local filename="$(basename "${1}")"
+    # If an epoch is listed, return it. Otherwise, assume an epoch of 0.
+    case "${filename}" in
+        *'!'*)
+            echo "${parts[0]#\!*}"
```
Ichimonji10: Whoops! This should be a reference to `filename`, not `parts`.

elyezer: Whoops! That was hiding very well when I read the code. Thanks for fixing it.

elyezer:
LGTM

PR #17, opened 2016-07-26
Generate signed and unsigned rpm fixtures.

emmawu:
including,
  unsigned rpm fixtures
  signed srpm fixtures
  unsigned srpm fixtures
  unsigned drpm fixtures


README.rst
```diff
@@ -50,6 +56,12 @@ are omitted.
 ``fixtures/rpm-invalid-updateinfo``
     See ``fixtures/rpm``.
 
+``fixtures/srpm``
+    The ``createrepo`` and ``modifyrepo`` utilities must be available.
+
+``fixtures/srpm-unsigned``
+    The ``createrepo``, ``modifyrepo`` and ``rpmsign`` utilities must be available.
```
Ichimonji10: There is no such utility as `rpmsign`. Do you mean `rpm-sign`?

Makefile
```diff
@@ -20,21 +28,40 @@ all: fixtures
 	$(warning The `all` target is deprecated. Use `fixtures` instead.)
 
 fixtures: fixtures/docker \
+	fixtures/drpm-unsigned \
```
Ichimonji10: Nitpick: all of the other lines below use spaces, and using a tab character here is out-of-place.

README.rst
```diff
@@ -34,9 +34,15 @@ are omitted.
...
     newgrp``.
 
+``fixtures/drpm-unsigned``
+    The ``createrepo`` and ``makedeltarpm`` utilities must be available.
+
 ``fixtures/rpm``
     The ``createrepo`` and ``modifyrepo`` utilities must be available.
 
+``fixtures/rpm-unsigned``
+    The ``createrepo``, ``modifyrepo`` and ``rpm-sign`` utilities must be available.
```
Ichimonji10: There is no executable named `rpm-sign`.  However, there is an executable named `rpmsign`. Do you mean that? (Confusingly, the `rpmsign` utility is provided by the `rpm-sign` package.)

rpm/del-fixtures-sign.sh
```diff
@@ -0,0 +1,44 @@
...
+    echo 1>&2 'Error: Too many arguments received.'
+    echo 1>&2
+    show_help 1>&2
+    exit 1
+else
+    target_dir="$(realpath --canonicalize-missing "${1}")"
+fi
+
+# Delete signatures from packages in target_dir recursively
+rpmsign --delsign `find ${target_dir} -print | grep -i '.*[.][sd]*rpm'`
```
Ichimonji10: Do not use backticks. They're deprecated in favor of `$()`, not least because they cannot be nested.

rpm/del-fixtures-sign.sh
```diff
@@ -0,0 +1,44 @@
...
+
+# Print a message to stdout explaining how to use this script.
+#
+# NOTE: $0 corresponds to the the script's name only in some shells and in some
+# contexts.
+show_help() {
+fmt <<EOF
+usage: del-fixtures-sign.sh <target_dir>
+
+Delete signatures of all packages recursively in <target_dir>
```
Ichimonji10: What are the "all packages" referred to by this sentence? All files ending in `.rpm`, `.drpm` and `.srpm`? Please clarify.

rpm/gen-fixtures-delta.sh
```diff
@@ -0,0 +1,160 @@
...
+	echo 1>&2 'Error: Not found enought RPM packages to DRPMS. At least 2 RPMS is required'
+	exit 1
+fi
+
+# Make DRPMS from RPMS
+# DRPMS is generated from all neighbouring RPMS in filtered RPMS list
+for (( i=0; i<${rpms_len}-1; i++ ));
+do
+	old_package=${found_rpms[i]}
+	new_package=${found_rpms[i+1]}
```
Ichimonji10: Tabs are pure evil. :wink:

rpm/gen-fixtures-delta.sh
```diff
@@ -0,0 +1,160 @@
...
+EOF
+}
+
+# Given a path to an RPM package, echo the package name.
+#
+# e.g. assets/walrus-baby-5.21-1.noarch.rpm → walrus-baby
+get_rpm_name() {
+    local filename="$(basename "${1}")"
+    IFS=$'\n' local parts=($(echo "${filename}" | tr - "\n"))
+    local end=`expr ${#parts[@]} - 2`
```
Ichimonji10: Backticks are deprecated and not part of the POSIX standard.

rpm/gen-fixtures-delta.sh
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/env sh
+#
+# Generate RPM repository with DRPM fixture data. Several RPM utilities must be installed,
+# including `createrepo` and `makedeltarpm`.
+#
+set -euo pipefail
+
+#------------------------------------------------------------------------------
+# Helper Functions
```
Ichimonji10: Nitpick: either drop this headline, or add a "business logic" headline below. I'd advocate for the former.

rpm/gen-fixtures-delta.sh
```diff
@@ -0,0 +1,160 @@
...
+found_rpms=()
+for (( i=0; i<${#rpms[@]}; i++ ));
+do
+    current_package_name="$(get_rpm_name "${rpms[i]}")"
+    if [ "${current_package_name}" == "${package_name}" ]; then
+        found_rpms+=("${rpms[i]}")
+    fi
+done
+
+readonly num_needed=2
```
Ichimonji10: This is a single-use variable. Interpolate its value into the code below and drop this variable, or use it in the string error message.

rpm/gen-fixtures-delta.sh
```diff
@@ -0,0 +1,160 @@
...
+
+	old_version="$(get_rpm_version "${old_package}")"
+	new_version="$(get_rpm_version "${new_package}" )"
+
+	old_release="$(get_rpm_release "${old_package}" "${old_arch}")"
+	new_release="$(get_rpm_release "${new_package}" "${new_arch}")"
+
+	delta_package="${new_name}-${old_version}-${old_release}_${new_version}-${new_release}.${new_arch}.drpm"
+
+	makedeltarpm ${old_package} ${new_package} "${output_dir}"/drpms/"${delta_package}"
```
Ichimonji10: Quote all variables to prevent word splitting issues. Also, limit lines to 80 characters wide if reasonably possible.

rpm/gen-fixtures-delta.sh
```diff
@@ -0,0 +1,160 @@
...
+    old_name="$(get_rpm_name "${old_package}")"
+    new_name="$(get_rpm_name "${old_package}")"
+
+    if [ "${old_name}" != "${new_name}" ]; then
+        echo 1>&2 'Error: Packages need has same name'
+        exit
+    fi
+
+	old_arch=$(get_rpm_arch "${old_package}")
+	new_arch=$(get_rpm_arch "${new_package}")
```
Ichimonji10: Quote all variable expansions to prevent word splitting issues.

rpm/gen-fixtures-delta.sh
```diff
@@ -0,0 +1,160 @@
...
+	old_release="$(get_rpm_release "${old_package}" "${old_arch}")"
+	new_release="$(get_rpm_release "${new_package}" "${new_arch}")"
+
+	delta_package="${new_name}-${old_version}-${old_release}_${new_version}-${new_release}.${new_arch}.drpm"
+
+	makedeltarpm ${old_package} ${new_package} "${output_dir}"/drpms/"${delta_package}"
+done
+
+# Generate DRPMS fixtures from RPMS and DRPMS
+createrepo --checksum sha256 --delta ${output_dir}
```
Ichimonji10: Quote all variable expansions to prevent word splitting issues. (Try cloning all of your repositories into a directory named `my code`, sometime, and see how many things break. It's astounding.)

rpm/gen-fixtures-delta.sh
```diff
@@ -0,0 +1,160 @@
...
+    IFS=$'\n' local parts=($(echo "${filename}" | tr . "\n"))
+    echo "${parts[-2]}"
+}
+
+#------------------------------------------------------------------------------
+# Business Logic
+#------------------------------------------------------------------------------
+
+# Fetch output_dir from user.
+if [ "$#" -lt 2 ]; then
```
Ichimonji10: So... it's valid to provide two arguments, yet `package_name="${3}"` below?

Ichimonji10:
This generally looks pretty good. However, my one remaining concern is the size of the yum DRPM files. At 1MB each, they add about 10MB to the size of the repository. That's OK, but not ideal, simply because of how much time it adds to repository clone times, and it'll compound if we ever decide to use different RPMs. Will that add another 10MB?

I should point out that I have a personal stake in this. I sometimes work from locations with slow connections that run at 1 Mbit/s or less.

README.rst
emmawu: I was thinking whether I should put package name or utility name there.. Later I found it should be the utility name, and updated descriptions in .sh files. But I forgot to change the name in README as well :(

rpm/del-fixtures-sign.sh
emmawu: "all packages" means all rpm/srpm/drpm files. Updated the description.

rpm/gen-fixtures-delta.sh
emmawu: Thanks. Added the variable to the error message.

rpm/gen-fixtures-delta.sh
emmawu: Thanks for reminding. I have updated and tested the script in a directory of which full path has spaces..It should be OK now.

README.rst
Ichimonji10: :+1:

Yeah, it's an interesting choice to make. In my opinion, it's better to write the utility name, because that tends to be more stable across different distributions and through time.

rpm/del-fixtures-sign.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env sh
```
Ichimonji10: Please declare the script as being a bash script, not a sh script. Some systems will ignore the request and start a bash shell anyway, but that's not something to count on.

rpm/del-fixtures-sign.sh
```diff
@@ -0,0 +1,41 @@
...
+    echo 1>&2 'Error: Too many arguments received.'
+    echo 1>&2
+    show_help 1>&2
+    exit 1
+else
+    target_dir="$(realpath --canonicalize-missing "${1}")"
+fi
+
+# Delete signatures from packages in target_dir recursively
+find "${target_dir}" -name "*rpm" -exec rpmsign --delsign {} \;
```
Ichimonji10: `find`'s `-exec` argument frequently masks non-zero return values. Use `xargs` instead. It's better at complaining when thing go wrong. Also, `*rpm` will catch files such as `not_an_rpm`. Use `\( -name 'pattern1' -o -name 'pattern2' \)`.

```bash
find "${target_dir}" \( -name '*.rpm' -o -name '*.drpm' -o -name '*.srpm' \) \
    -print0 | xargs -0 rpmsign --delsign
```

rpm/gen-fixtures-delta.sh
```diff
@@ -0,0 +1,154 @@
+#!/usr/bin/env sh
```
Ichimonji10: Declare this script as a bash script, not an sh script.

rpm/gen-fixtures-delta.sh
```diff
@@ -0,0 +1,154 @@
...
+    current_package_name="$(get_rpm_name "${rpms[i]}")"
+    if [ "${current_package_name}" == "${package_name}" ]; then
+        found_rpms+=("${rpms[i]}")
+    fi
+done
+
+readonly num_needed=2
+rpms_len=${#found_rpms[@]}
+if [ ${rpms_len}  -lt ${num_needed} ]; then
+    echo 1>&2 'Error: Need at least ${num_needed} RPMS. But found ${rpms_len}'
```
Ichimonji10: Variables are interpolated into double-quoted strings, but not single-quoted strings. This will literally print the following message:

    Error: Need at least ${num_needed} RPMS. But found ${rpms_len}

rpm/del-fixtures-sign.sh
```diff
@@ -0,0 +1,41 @@
...
+    echo 1>&2
+    show_help 1>&2
+    exit 1
+elif [ "$#" -gt 1 ]; then
+    echo 1>&2 'Error: Too many arguments received.'
+    echo 1>&2
+    show_help 1>&2
+    exit 1
+else
+    target_dir="$(realpath --canonicalize-missing "${1}")"
```
Ichimonji10: `--canonicalize-missing` is an inappropriate option to use here. From `realpath`'s man page:

```
       Print the resolved absolute file name; all but the last component must exist

       -e, --canonicalize-existing
              all components of the path must exist

       -m, --canonicalize-missing
              no path components need exist or be a directory
```

`--canonicalize-existing` should be used.

rpm/gen-fixtures-delta.sh
```diff
@@ -0,0 +1,154 @@
...
+EOF
+}
+
+# Given a path to an RPM package, echo the package name.
+#
+# e.g. assets/walrus-baby-5.21-1.noarch.rpm → walrus-baby
+get_rpm_name() {
+    local filename="$(basename "${1}")"
+    IFS=$'\n' local parts=($(echo "${filename}" | tr - "\n"))
+    local end=$(expr ${#parts[@]} - 2)
```
Ichimonji10: `expr` is ancient. Use `$(( ))` for math.

```bash
end=$(( ${#parts[@]} - 2 ))
```

rpm/gen-fixtures-delta.sh
```diff
@@ -0,0 +1,154 @@
...
+    show_help 1>&2
+    exit 1
+elif [ "$#" -gt 3 ]; then
+    echo 1>&2 'Error: Too many arguments received.'
+    echo 1>&2
+    show_help 1>&2
+    exit 1
+else
+    output_dir="$(realpath --canonicalize-missing "${1}")"
+    assets_dir="$(realpath "${2}")"
```
Ichimonji10: Consider using `--canonicalize-existing`.

rpm/gen-fixtures-delta.sh
```diff
@@ -0,0 +1,154 @@
...
+    if [ "${name_1}" != "${name_2}" ]; then
+        echo 1>&2 'Error: Packages need has same name'
+        exit
+    fi
+
+    arch_1="$(get_rpm_arch "${pkg_1}")"
+    arch_2="$(get_rpm_arch "${pkg_2}")"
+
+    if [ "${arch_1}" != "${arch_2}" ]; then
+        echo 1>&2 'Error: Old and new packages need has same arch'
```
Ichimonji10: Does the `makedeltarpm` tool perform this check? If so, let's omit this check and let that tool do error checking for us. If not, how about print this message instead?

```bash
    arch_1="$(get_rpm_arch "${pkg_1}")"
    arch_2="$(get_rpm_arch "${pkg_2}")"
    if [ "${arch_1}" != "${arch_2}" ]; then
        fmt 1>&2 <<EOF
Error: Old and new packages must have the same architecture, but are different.
Package architectures are ${arch_1} and ${arch_2}.
EOF
        exit 1
    fi
```

rpm/gen-fixtures-delta.sh
```diff
@@ -0,0 +1,154 @@
...
+for (( i=0; i<${rpms_len}-1; i++ ));
+do
+    pkg_1="${found_rpms[i]}"
+    pkg_2="${found_rpms[i+1]}"
+
+    name_1="$(get_rpm_name "${pkg_1}")"
+    name_2="$(get_rpm_name "${pkg_2}")"
+
+    if [ "${name_1}" != "${name_2}" ]; then
+        echo 1>&2 'Error: Packages need has same name'
```
Ichimonji10: Does the `makedeltarpm` tool perform this check? If so, let's omit this check and let that tool do error checking for us. If not, how about printing this message instead?

```bash
    name_1="$(get_rpm_name "${pkg_1}")"
    name_2="$(get_rpm_name "${pkg_2}")"
    if [ "${name_1}" != "${name_2}" ]; then
        fmt 1>&2 <<EOF
Error: Old and new packages must have the same name, but are different. Package
names are ${name_1} and ${name_2}.
EOF
        exit 1
    fi
```

rpm/gen-fixtures-delta.sh
```diff
@@ -0,0 +1,154 @@
...
+    package_name="${3}"
+fi
+
+# Copy all RPMs from assets_dir to output_dir
+# Create directory 'drpms' for DRPMS
+mkdir -vp "${output_dir}"/drpms
+cp -t "${output_dir}" "${assets_dir}/"*.rpm
+
+# Filter and sort all required RPMs.
+rpms="${output_dir}"/"${package_name}"*.rpm
```
Ichimonji10: This will literally assign the string `${output_dir}"/"${package_name}"*.rpm` to the `rpms` variable. Here's a simple example:

```bash
touch myfoo mybar
var=my*
echo "$var"  # my*
echo $var  # myfoo myvar
```

Also, glob expansion produces an array, so it'd be good to use array syntax if an array is expected.

Nitpick: A benefit of the `${...}` variable name syntax is that it's easy to find variables even in complicated strings. In my opinion, `"${foo}/${bar}"` is easier to read than `"${foo}"/"${bar}"`.

Try this instead:

```bash
rpms=( "$output_dir}/${package_name}"*.rpm )
```

rpm/gen-fixtures-delta.sh
```diff
@@ -0,0 +1,154 @@
...
+do
+    current_package_name="$(get_rpm_name "${rpms[i]}")"
+    if [ "${current_package_name}" == "${package_name}" ]; then
+        found_rpms+=("${rpms[i]}")
+    fi
+done
+
+readonly num_needed=2
+rpms_len=${#found_rpms[@]}
+if [ ${rpms_len}  -lt ${num_needed} ]; then
```
Ichimonji10: Use double quotes to prevent word expansion issues.

rpm/gen-fixtures-delta.sh
```diff
@@ -0,0 +1,154 @@
...
+    echo 1>&2 'Error: Need at least ${num_needed} RPMS. But found ${rpms_len}'
+    exit 1
+fi
+
+# Make DRPMS from RPMS
+# DRPMS is generated from all neighbouring RPMS in filtered RPMS list
+for (( i=0; i<${rpms_len}-1; i++ ));
```
Ichimonji10: This can be written as:

```bash
for (( i=0; i < rpms_len - 1; i++ )); do
```

rpm/gen-fixtures-delta.sh
```diff
@@ -0,0 +1,154 @@
...
+}
+
+# Given a path to an RPM package, echo the package name.
+#
+# e.g. assets/walrus-baby-5.21-1.noarch.rpm → walrus-baby
+get_rpm_name() {
+    local filename="$(basename "${1}")"
+    IFS=$'\n' local parts=($(echo "${filename}" | tr - "\n"))
+    local end=$(expr ${#parts[@]} - 2)
+    for ((i=0; i<${end}; i++)); do
```
Ichimonji10: This can be written as:

```bash
for (( i=0; i < end; i++ )); do
```

rpm/gen-fixtures-delta.sh
```diff
@@ -0,0 +1,154 @@
...
+    rpm_name="${rpm_name::-1}"
+    echo "${rpm_name#*\!}"  # strip epoch
+}
+
+# Given a path to an RPM package, echo the package version.
+#
+# e.g. assets/walrus-5.21-1.noarch.rpm → 5.21
+get_rpm_version() {
+    local filename="$(basename "${1}")"
+    IFS=$'\n' local parts=($(echo "${filename}" | tr - "\n"))
```
Ichimonji10: I *just* barely learned about the value of declaring variables separately from using them. You can see an example of the dangers [here](http://pastebin.com/m9VPQ1KS). In short, write this instead:

```bash
    local filename parts
    filename="$(basename "${1}")"
    IFS=$'\n' parts=($(echo "${filename}" | tr - "\n"))
```

Ichimonji10:
Many of my comments come from [shellcheck](https://www.shellcheck.net/). You can use shellcheck by installing it using your package manager and executing a command like `find . -name '*sh' -print0 | xargs -0 shellcheck`. This is, of course, a pain to type out. A better option is to rebase your development branch onto the upstream/master branch. Some [improvements](https://github.com/PulpQE/pulp-fixtures/commits/master) have been made, such as the addition of the `lint` make target.

Also, if you know how to squash commits, please do so. Otherwise, the reviewer who merges your work will do so. Beware that if the reviewer squashes your commits for you, they will be forced to come up with an original commit message, and it may or may not be to your liking.

Given the simple commit history at play, it's possible to rebase and squash your changes with the following script:

```bash
#!/usr/bin/env bash
set -euo pipefail

# Clone the existing repository and verify that it works.
git clone https://github.com/emmawu/pulp-fixtures.git
cd pulp-fixtures
git checkout develop
make fixtures/{drpm-unsigned,rpm-unsigned,srpm,srpm-unsigned}

# Add a remote tracking branch named "upstream".
git remote add upstream https://github.com/PulpQE/pulp-fixtures.git
git fetch --all

# Rebase the "develop" branch, and squash its commits.
git branch develop-old  # make a backup branch
git rebase upstream/master
git rebase --interactive upstream/master develop

# Verify that the updated code still works, and that the commit history looks
# sane.
make clean
make fixtures/{drpm-unsigned,rpm-unsigned,srpm,srpm-unsigned}
git log --decorate --graph --pretty=oneline --abbrev-commit --all

# If you're confident in the changes, push them.
# git push --force origin develop
```

By the way, notice how the `--force` option is used when pushing the branch? This is because we're *intentionally* throwing away commits. Git is very good about keeping commits around and warning you when commits may be lost.

Also, a big **thank you** for shrinking down the size of the binary blobs. I'm very, very happy to see that. :smile:

Finally, here's what you'll want to do when executing `git rebase --interactive upstream/master`. There's instructions right in the text editor, but you may appreciate the example.

screenshot_2016-07-27_15-48-57

rpm/gen-fixtures-delta.sh
emmawu: I think this one should use option '--canonicalize-missing'. Unlike del-fixtures-sign.sh,  ${output_dir}" is not expected to exist before running gen-fixtures-delta.sh. It's removed by `make clean`. And gen-fixtures-delta.sh will create this directory ${output_dir}. 
This is also true to other scripts except del-fixture-sign.sh

rpm/gen-fixtures-delta.sh
emmawu: makedeltarpm doesn't do the check. It will take whatever rpm packages and generate the drpm package in the name specified in the 3rd argument.
Thanks. I have updated the error message.

emmawu:
Thank you, @Ichimonji10. The instructions you've given for shellcheck and commit squashing is very helpful.
I have squashed my commits and reabsed it on upstream/master. And also updated my scripts based on all of your comments and shellcheck complaint.  
Please review.

rpm/gen-fixtures-delta.sh
Ichimonji10: Yes, `$output_dir` should use `--canonicalize-existing`. But I'm looking at `$assets_dir`.

rpm/gen-fixtures-delta.sh
Ichimonji10: Thanks for the clarification. :smile:

Ichimonji10:
My only significant suggestion is that the `createrepo` command use the `--deltas` option, not the `--delta` option. The man pages for createrepo and createrepo_c both make mention of `--deltas`, not `--delta`. I'll do this in a follow-up commit. I've tested out the change, and nothing changes. I assume that `--delta` is supported for legacy reasons.

Ichimonji10:
Merged, with a follow-up commit of 8a003e048dfa814301a66734d3a022a972821fe1. Thank you!

PR #21, opened 2016-07-27
Add .travis.yml

elyezer:
Run `make lint` for running shellcheck on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,14 @@
+sudo: required
+dist: trusty
+before_install:
+    - sudo apt-get -qq update
+    - sudo apt-get install -y cabal-install
+    - cabal update
+    - git clone https://github.com/koalaman/shellcheck.git
+    - cd shellcheck
+    - cabal install
+    - cd ..
+    - export PATH="$HOME/.cabal/bin:$PATH"
+    - which shellcheck
+script:
+    make lint


elyezer:
Travis Ubuntu versions does not have shellcheck on its repositories. We will need to create a Docker image in order to run the lint. Using Docker will be also good to generate the fixtures and see if they will not produce any errors.

Ichimonji10:
As an idea, one can also install packages from source in Travis.

elyezer:
@Ichimonji10 compiling from source worked! I have adjusted the `.travis.yml` file to compile shellcheck and also opened this issue https://github.com/koalaman/shellcheck/issues/721. If we get the package built for the OS versions Travis supports then that will make the build much faster.

This PR is finally ready to be reviewed.

.travis.yml
```diff
@@ -0,0 +1,14 @@
...
+before_install:
+    - sudo apt-get -qq update
+    - sudo apt-get install -y cabal-install
+    - cabal update
+    - git clone https://github.com/koalaman/shellcheck.git
+    - cd shellcheck
+    - cabal install
+    - cd ..
+    - export PATH="$HOME/.cabal/bin:$PATH"
+    - which shellcheck
```
elyezer: This is to ensure shellcheck is properly setup on the PATH

Ichimonji10:
LGTM

Interestingly, I had to install the Glasgow Haskell Compiler via the "ghc" package. Without it, the `cabal install` command returned an error. But, I'm not on Ubuntu, or any other Debian-based distribution for that matter, and Travis seems happy.

PR #22, opened 2016-07-27
Add json api to python fixture

asmacdo:
Mimics the json api of PyPI. The actual structure and files are produced
by a pulp publish so this effectively tests a pulp to pulp sync as well.

### DIFF
--- a/python/assets/pypi/shelf-reader/json/index.html
+++ b/python/assets/pypi/shelf-reader/json/index.html
@@ -1,0 +1,27 @@
+{
+  "info": {
+    "author": "Austin Macdonald",
+    "name": "shelf-reader",
+    "summary": "Make sure your collections are in call number order."
+  },
+  "releases": {
+    "0.1": [
+      {
+        "md5_digest": "2dac570a33d88ca224be86759be59376",
+        "filename": "shelf-reader-0.1.tar.gz",
+        "packagetype": "sdist",
+        "checksum_type": "sha512",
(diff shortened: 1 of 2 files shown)


Ichimonji10:
PyPi provides this JSON metadata as a pair of identical files, one ending in `.json` and the other in `.html`. See:

* https://pypi.python.org/pypi/shelf-reader/json/index.json
* https://pypi.python.org/pypi/pulp-smash/json/index.html

If the intent is to mimic PyPi, does it make more sense to create an `index.json` file, and create a symlink pointing to it named `index.html`?

On a nit-pickier note, `json_reformat` is good at making files look nice.

Here's a code summary of my suggestions:

```bash
#!/usr/bin/env bash
set -euo pipefail

json_reformat \
    < python/assets/pypi/shelf-reader/json/index.html \
    > python/assets/pypi/shelf-reader/json/index.json
rm python/assets/pypi/shelf-reader/json/index.html
ln -s  ./index.json python/assets/pypi/shelf-reader/json/index.html
```

asmacdo:
Unfortunately, the symlink didn't work the way I expected with default `python -m http.server`, so I just created a second file.

Ichimonji10:
This change looks sane. I also make sure to generate the updated fixtures, host them with `python3 -m http.server`, update the Pulp Smash constants file to point to the served files, and run `python -m unittest discover pulp_smash.tests.python`. All tests passed. Merging. Thank you, @asmacdo!

elyezer:
:+1:

PR #24, opened 2016-08-05
Sign all RPMs with Pulp QE GPG key

elyezer:
Also publish the Pulp QE public GPG key.


seandst:
Is it okay that the private key isn't also included? Without it, future fixture-makers would be unable to sign packages with the correct key

elyezer:
@seandst I have not included the private key since I don't know about having it published would be something bad. I think it will not since the purpose is just for testing and it should be imported and used on testing machines.

If we include the private key then we can improve the fixtures generator scripts to also import the private key and use it to sign the packages when generating or create a separated script that can update the signing to the Pulp QE key.

Ichimonji10:
How were the packages signed? With `rpmsign --resign ${packages[*]}`?

Ichimonji10:
Also, what's the motivation for this change? Is it so that issues like https://github.com/PulpQE/pulp-smash/issues/347 may be taken care of?

elyezer:
@Ichimonji10 the RPM were resigned using `rpm --addsign "${packages[*]}"`. I tried with `--addsign` and it worked, not sure if it does something different from the `--resign`.

The main goal of this change is to have control of the expected signing ID, right now we have different signing keys. Having a normalized ID is better because we can ensure that all RPMs will be signed using the same key.

Also the other point is that if we should include the private key as well so everyone can sign the packages using the same key. By making the private key public we should add a big warning saying that it is not secure.

Ichimonji10:
The difference is that `--addsign` adds an additional signature to the RPMs, whereas `--resign` replaces the signatures on the RPMs. It's even possible to add a single signature multiple times to a single package by repeatedly using `--addsign`.

Ichimonji10:
[This page](http://www.rpm.org/max-rpm/s1-rpm-pgp-signing-packages.html) is where I'm getting my information from. :smile:

I think it makes sense to put the private signing key out in the open by adding it to this repository. This repository is not supposed to be an authoritative source of any packages at all. And yes, let's put warnings where appropriate.

> The main goal of this change is to have control of the expected signing ID, right now we have different signing keys. Having a normalized ID is better because we can ensure that all RPMs will be signed using the same key.

Makes sense. I've idly wondered about how the packages are signed in the past, even before we had test cases that made specific use of the keys.

elyezer:
Marking the PR as needs works since it seems that we agreed on:

1. Adding the private key
2. Removing all but the Pulp QE signature from the RPMs
3. Adding a big warning about the private key being public available

I am going to work on the changes and will update here when I'm done.

Ichimonji10:
Sounds great to me. A possible future improvement is to make all of the RPMs unsigned, and to sign them as part of the fixture generation process. But that change might require some changes to the fixture generation scripts, and I'm perfectly happy to see the improvement listed above.

elyezer:
I think the best option is to have all signed and then remove the signature, like is done currently. That's because in order to sign you will need to:

* Have gpg and rpm-sign installed
* Configure `~/.rpmmacros` to use the Pulp QE key
* Add the private key to the GPG in order to use it during the signature process
* Run the `rpm --addsign` to sign the RPMs.

With that said, I am not sure if everyone will like to edit `~/.rpmmacros`, add a public available GPG key to its key chain.

We can provide documentation on how to have that setup so, if new RPMs are needed, they can be signed properly.

Ichimonji10:
Ahhh, you're right. That's a real pain to do. Let's keep the barrier to entry low.

elyezer:
Update:

* Removed current signature from all RPMs using `find . \( -name '*.rpm' -o -name '*.srpm' -o -name '*.drpm' \) | xargs rpmsign --delsign`
* Check if all RPMs had not signature using `find . \( -name '*.rpm' -o -name '*.srpm' -o -name '*.drpm' \) | xargs rpm --checksig`
* Signed all RPMs using the Pulp QE key using `find . \( -name '*.rpm' -o -name '*.srpm' -o -name '*.drpm' \) | xargs rpm --addsign` and check if the signature as added using the `--checksig`.
* Added the Pulp QE private GPG key

Missing improving the README with information about how to setup in order to sign the RPMs and the big warning about the private key.

Ichimonji10:
Would this help?

```diff
diff --git a/README.rst b/README.rst
index 30e23bc..816fd4e 100644
--- a/README.rst
+++ b/README.rst
@@ -26,6 +26,13 @@ ensure the necessary utilities are available and usable. Dependencies are listed
 below, according to make target. Common system utilities like ``fmt``, ``patch``
 and ``realpath`` are omitted.
 
+All RPM, SRPM and DRPM fixtures are signed with an OpenPGP-compatible keypair,
+except where otherwise stated.
+
+.. WARNING:: The private key used to sign RPM, SRPM and DRPM files is publicly
+    available. The signatures on these files afford absolutely *no* security.
+    The signatures are present only for testing purposes.
+
 ``lint``
     The ``shellcheck`` executable must be available.
 
@@ -71,6 +78,18 @@ and ``realpath`` are omitted.
 ``fixtures/srpm-unsigned``
     The ``createrepo``, ``modifyrepo`` and ``rpmsign`` utilities must be available.
 
+Package Signatures
+------------------
+
+The RPM, SRPM and DRPM assets are signed, and signatures are stripped as needed
+when generating fixtures. The opposite approach of using unsigned fixtures and
+signing them as needed is safer, as it better prevents the keypair from going
+out of sync with fixtures. However, the sign-as-needed approach is not taken
+because installing and configuring GnuPG can be a hassle.
+
+Should any new assets need to be signed, or should any existing assets need to
+be re-signed, a brief guide to doing so follows.
+
 .. _Pulp #2020: https://pulp.plan.io/issues/2020
 .. _Pulp RPM Errata:
     https://docs.pulpproject.org/plugins/pulp_rpm/tech-reference/yum-plugins.html#errata
```

Ichimonji10:
Beware that the diff above doesn't include an actual guide to configuring GnuPG and/or stripping/checking/signing files. It just includes prose.

elyezer:
I have updated the PR with the suggested diff. Will try to add more detailed information for the signature process.

Ichimonji10:
This is great. What do you think about moving the two lines of bash code into their own code block, and writing `find rpm` instead of `find path/to/packages`?

```diff
diff --git a/README.rst b/README.rst
index e00f32d..1c187fe 100644
--- a/README.rst
+++ b/README.rst
@@ -99,19 +99,20 @@ with the following lines::
     %_gpg_name  Pulp QE
 
 That will ensure the RPM sign is going to use the Pulp QE key imported
-previously. Finally, sign the packages by running ``find path/to/packages \(
--name '*.rpm' -o -name '*.srpm' -o -name '*.drpm' \) | xargs rpm --addsign``.
+previously. Finally, sign the packages and verify their signatures with:
 
-You can run ``find path/to/packages \( -name '*.rpm' -o -name '*.srpm' -o -name
-'*.drpm' \) | xargs rpm --checksig`` to check if the packages were signed
-properly. If the package is not signed then an output line like
+.. code-block:: sh
+
+    find rpm \( -name '*.rpm' -o -name '*.srpm' -o -name '*.drpm' \) | xargs rpm --addsign
+    find rpm \( -name '*.rpm' -o -name '*.srpm' -o -name '*.drpm' \) | xargs rpm --checksig
+
+If the package is not signed then an output line like
 ``fixtures/rpm-unsigned/lion-0.4-1.noarch.rpm: sha1 md5 OK`` will be shown, on
 the other hand, if the package is signed, an output line like
-``fixtures/rpm/lion-0.4-1.noarch.rpm: (RSA) sha1 ((MD5) PGP) md5 NOT OK
-(MISSING KEYS: RSA#269d9d98 (MD5) PGP#269d9d98)``. Be aware that the ``MISSING
-KEYS`` message is present because the public key was not imported to the RPM
-keys store and probably you will not want to import it since the key is not
-safe.
+``fixtures/rpm/lion-0.4-1.noarch.rpm: (RSA) sha1 ((MD5) PGP) md5 NOT OK (MISSING
+KEYS: RSA#269d9d98 (MD5) PGP#269d9d98)``. Be aware that the ``MISSING KEYS``
+message is present because the public key was not imported to the RPM keys
+store, and you probably will not want to import it since the key is not safe.
 
 .. _Pulp #2020: https://pulp.plan.io/issues/2020
 .. _Pulp RPM Errata:
```

Ichimonji10:
Merging. The build failure was due to an issue with package compilation.

Ichimonji10:
Rebased and merged as c1fe92ab252cec73d2e3bc018524cabb51603251. Thank you, @elyezer!

PR #28, opened 2016-09-15
Change how package signing is handled

Ichimonji10:
Without this change, the readme reads as follows:

> The RPM, SRPM and DRPM assets are signed, and signatures are stripped
> as needed when generating fixtures. The opposite approach of using
> unsigned fixtures and signing them as needed is safer, as it better
> prevents the keypair from going out of sync with fixtures. However,
> the sign-as-needed approach is not taken because installing and
> configuring GnuPG can be a hassle.

With this change, the readme reads as follows:

> The RPM, SRPM and DRPM assets are unsigned, and signatures are added
> as needed when generating fixtures. The opposite approach of using
> signed assets and stripping signatures as needed is less safe, as the
> keypair can more easily go out of sync with the assets.

Critically, user files such as `~/.gnupg` and `~/.rpmmacros` are *not*
touched by Pulp Fixtures.

The `gpg` executable is used by make targets that perform package
signing. This is a new dependency.


Ichimonji10:
I need to drop checksums from `pulp_smash.constants` before this can be merged.

Makefile
```diff
@@ -64,16 +68,21 @@ fixtures/rpm-invalid-updateinfo:
...
 
 fixtures/rpm-updated-updateinfo:
 	rpm/gen-patched-fixtures.sh $@ rpm/updated-updateinfo.patch
 
-fixtures/srpm:
+fixtures/srpm: gnupghome
 	rpm/gen-fixtures.sh $@ rpm/assets-srpm
+	GNUPGHOME=$$(realpath -e gnupghome) rpmsign --define '_gpg_name Pulp QE' \
+		--addsign --fskpath ./rpm/GPG-RPM-PRIVATE-KEY-pulp-qe --signfiles \
+		$$(find $@ -name '*.src.rpm')
```
Ichimonji10: These three lines are almost copied and pasted from above. This is an area that's ripe for a future refactor.

Ichimonji10:
Rebased.

elyezer:
LGTM

Ichimonji10:
Thanks for the reviews, @elyezer. I've merged https://github.com/PulpQE/pulp-smash/pull/395, so this should now be a candidate for merging.

To make sure, I ran the Jenkins job to generate fixtures and executed `python -m unittest pulp_smash.tests.rpm.api_v2.test_download_policies.OnDemandTestCase`. This exercised one of the just-modified tests, and all went well. If we merge this PR and something breaks, we can point to this commit and say it's at fault.

PR #42, opened 2016-10-13
Update to use newly published Pulp metatdata

asmacdo:
Also fixes a sync problem.

### DIFF
--- a/python/assets/pypi/shelf-reader/json/index.html
+++ b/python/assets/pypi/shelf-reader/json/index.html
@@ -1,27 +1,27 @@
 {
-  "info": {
-    "author": "Austin Macdonald",
-    "name": "shelf-reader",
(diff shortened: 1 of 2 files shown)


Ichimonji10:
The `path` is the only thing that's changed, right? Everything appears to be a semantic no-op, like rearranging keys in a dict or adding additional indentation.

asmacdo:
@Ichimonji10 correct.

Ichimonji10:
Tested against Pulp 2.8 and 2.11. All Python API tests still pass. This change makes this fixture more like what PyPI hosts, so :+1: to that.

Ichimonji10:
Updated fixtures compiled and made available on repos.fedorapeople.org, and all Python tests run with a set of Pulp 2.8-2.11 systems. All tests passed.

PR #44, opened 2016-10-13
Add PyPI-compatible Python repository

Ichimonji10:
Add a new make target, `fixtures/python-pypi`. This make target is a
counterpart to `fixtures/python-pulp`. The documents in this fixture are
taken directly from PyPI, and only several URLs are modified.

Fix: https://github.com/PulpQE/pulp-fixtures/issues/40


Ichimonji10:
Elyezer is out on PTO for a while. I've manually tested this, and will test it again after merging.

PR #46, opened 2016-12-01
Give unbound variable a default value

Ichimonji10:
If `rpm/gen-fixtures.sh` is called with `--packages-dir
<relative-path>`, then an internal variable named `$packages_dir` is
set. If called without that flag, the internal `$packages_dir` variable
is not set, which leads to an unbound variable error. Fix this issue by
giving `$packages_dir` a default value for those cases where
`rpm/gen-fixtures.sh` is called without `--packages-dir`.

This change fixes targets such as `fixtures/rpm-unsigned`.


Ichimonji10:
The following command now succeeds:

    make clean fixtures/rpm-{unsigned,signed,alt-layout}

Ichimonji10:
Cabal failed to install again. :/ Locally, `make lint` passes.

PR #47, opened 2016-12-01
Use shellcheck 4.4

elyezer:
Shellcheck 4.5 is raising errors when installing, use 4.4 instead to make sure
Travis will produce sane results.


Ichimonji10:
See: https://github.com/koalaman/shellcheck/issues/778

PR #48, opened 2017-01-17
Add file fixtures

elyezer:
File fixture is a repository where Pulp will publish the files. On Pulp
2 it is imported by the iso command but it does not mean that the files
should be a real ISO, it can be any file type. On Pulp 3 it will be
renamed to file and that is why this is called file fixtures.

A valid file fixture repository must have the fixtures files and the
PULP_MANIFEST file, the latter is a CSV file where each line is the file
path relative to the fixtures URL, the sha256sum of the file and the
size, in bytes, of the file.

Also add the file-mixed fixture which will generate a file fixture with
valid and invalid entries on the PULP_MANIFEST file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,10 @@ help:
 	@echo "                  to create DRPM fixtures with signed packages"
 	@echo "  fixtures/drpm-unsigned"
 	@echo "                  to create DRPM fixtures with unsigned packages"
+	@echo "  fixtures/file   to create File fixtures"
+	@echo "  fixtures/file-mixed"
+	@echo "                  to create File fixtures with some not available"
+	@echo "                  files on the PULP_MANIFEST"
 	@echo "  fixtures/python-pulp"
 	@echo "                  to create a Pulp Python repository"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -65,6 +69,8 @@ fixtures: fixtures/docker \
 	fixtures/drpm \
 	fixtures/drpm-signed \
 	fixtures/drpm-unsigned \
+	fixtures/file \
+	fixtures/file-bad \
```
Ichimonji10: Should this read "file-bad" or "file-mixed" ?

Ichimonji10:
Just fix the one comment I've left. Otherwise, LGTM.

Makefile
elyezer: Good catch, I started with file-bad but since it has also valid entries decided to update to file-mixed which I think makes more sense. Will fix.

Ichimonji10:
Thanks!

elyezer:
Thank you @Ichimonji10 for the review. Will wait travis getting green and will merge.

PR #50, opened 2017-01-31
Add rpm-with-pulp-distribution fixture

elyezer:
This fixture creates a RPM repository with some extra files that are
mapped by the `PULP_DISTRIBUTION.xml` file


Makefile
```diff
@@ -89,6 +92,7 @@ fixtures: fixtures/docker \
 	fixtures/rpm \
 	fixtures/rpm-alt-layout \
 	fixtures/rpm-erratum \
+	fixtures/rpm-extra-repodata-files \
```
Ichimonji10: This is wrong.

PR #52, opened 2017-02-23
Add puppet fixtures

Ichimonji10:
Add a new make target, `fixtures/puppet`, which generates a dummy Puppet
module. The new make target depends on the ``puppet`` executable.

Fix: https://github.com/PulpQE/pulp-fixtures/issues/31

### DIFF
--- a/puppet/gen-module.sh
+++ b/puppet/gen-module.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+#
+# Generate a Puppet module.
+#
+set -euo pipefail
+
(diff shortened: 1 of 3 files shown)


puppet/gen-module.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+#
+# Generate an RPM repository.
```
elyezer: s/RPM repository/Puppet module?

Ichimonji10: Yes!

elyezer: That happens :smile, thanks for fixing.:

PR #56, opened 2017-07-19
Add script to generate ostree fixtures.

jortel:
Add script to generate ostree fixtures.  Specifically a small repository in `fixtures/ostree/small` used by some of the smash tests.

### DIFF
--- a/ostree/gen-fixtures.sh
+++ b/ostree/gen-fixtures.sh
@@ -1,0 +1,87 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+DIR=""
+TREE=""
+USAGE="
+Usage: gen-fixtures.sh <output-dir>
+
+Generate an OSTree repository with (2) branches:
+ - rawhide
+ - stable
+
+Each branch will have (3) commits for versions:
+ - 1.0
+ - 1.1
+ - 1.2
+
+The OSTree repository will be created in <output-dir>.
+The directory will be be created.
+"
+
+show_help()
+{
+  echo "${USAGE}"
+}
+
+_mkdir()
+{
+  if [ -e "${DIR}" ]; then
+    if [ ! -z "$(ls -A "${DIR}")" ]; then
+      echo 1>&2 ""
+      echo 1>&2 "${DIR} must be empty."
+      echo 1>&2 ""
+      exit 1
+    fi
+  else
+    mkdir -p "${DIR}"
+  fi
+}
+
+gen_tree()
+{
+  # The tree used for the ostree commits must be a directory
+  # containing at least (1) file.  Using this script for convenience.
+  TREE=$(mktemp -d /tmp/tree.XXXXX)
+  touch "${TREE}/fake.img"
+}
+
+gen_repository()
+{
+  ostree init --repo="${DIR}" --mode=archive-z2
+  # branch: rawhide
+  ostree commit --repo="${DIR}" -b rawhide --add-metadata-string=version=1.1 "${TREE}"
+  ostree commit --repo="${DIR}" -b rawhide --add-metadata-string=version=1.2 "${TREE}"
+  ostree commit --repo="${DIR}" -b rawhide --add-metadata-string=version=1.3 "${TREE}"
+  # branch: stable
+  ostree commit --repo="${DIR}" -b stable --add-metadata-string=version=1.1 "${TREE}"
+  ostree commit --repo="${DIR}" -b stable --add-metadata-string=version=1.2 "${TREE}"
+  ostree commit --repo="${DIR}" -b stable --add-metadata-string=version=1.3 "${TREE}"
+  # summary
+  ostree summary -u --repo="${DIR}"
+}
+
+clean()
+{
+  if [ -d "${TREE}" ]; then
+    rm -rf "${TREE}"
+  fi
+}
+
+
+if [ "$#" -ne "1" ]; then
+  show_help
+  exit 1
+fi
+
+DIR="$1"
+
+trap clean EXIT
+
+_mkdir
+gen_tree
+gen_repository
+clean
+echo "Created repository in: ${DIR}"
+
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ help:
 	@echo "  fixtures/file-mixed"
 	@echo "                  to create File fixtures with some not available"
 	@echo "                  files on the PULP_MANIFEST"
+	@echo "  fixtures/ostree to create a OSTree ostree repositories"
 	@echo "  fixtures/puppet to create a dummy Puppet module"
 	@echo "  fixtures/python-pulp"
 	@echo "                  to create a Pulp Python repository"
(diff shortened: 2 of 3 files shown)


ostree/gen-fixtures.sh
```diff
@@ -0,0 +1,81 @@
...
+      echo 1>&2 "${DIR} must be empty."
+      echo 1>&2 ""
+      exit 1
+    fi
+  else
+    mkdir -p ${DIR}
+  fi
+}
+
+gen_tree()
```
Ichimonji10: This function copies this script into `/tmp/tree.XXXXX/`. What function does that serve?

ostree/gen-fixtures.sh
```diff
@@ -0,0 +1,81 @@
...
+{
+  cp $0 ${TREE}
+}
+
+gen_repository()
+{
+  mkdir -p ${DIR}
```
Ichimonji10: `_mkdir()` already does this, right?

jortel:
@Ichimonji10, thanks for the review.  Ready for 2nd review.

ostree/gen-fixtures.sh
```diff
@@ -0,0 +1,80 @@
...
+  rm -rf "${TREE}"
+}
+
+
+if [ "$#" -ne "1" ]; then
+  show_help
+  exit 1
+fi
+
+DIR=$1
```
Ichimonji10: Please quote this. This will break if the script is called as `ostree/gen-fixtures.sh "fixtures/ostree/small tree"`.

ostree/gen-fixtures.sh
```diff
@@ -0,0 +1,80 @@
...
+{
+  ostree init --repo=${DIR} --mode=archive-z2
+  # branch: rawhide
+  ostree commit --repo=${DIR} -b rawhide --add-metadata-string=version=1.1 "${TREE}"
+  ostree commit --repo=${DIR} -b rawhide --add-metadata-string=version=1.2 "${TREE}"
+  ostree commit --repo=${DIR} -b rawhide --add-metadata-string=version=1.3 "${TREE}"
+  # branch: stable
+  ostree commit --repo=${DIR} -b stable --add-metadata-string=version=1.1 "${TREE}"
+  ostree commit --repo=${DIR} -b stable --add-metadata-string=version=1.2 "${TREE}"
+  ostree commit --repo=${DIR} -b stable --add-metadata-string=version=1.3 "${TREE}"
```
Ichimonji10: Please quote `${DIR}`. This will break if the script is called as `ostree/gen-fixtures.sh "fixtures/ostree/small tree"`.

ostree/gen-fixtures.sh
```diff
@@ -0,0 +1,80 @@
...
+
+show_help()
+{
+  echo "${USAGE}"
+}
+
+_mkdir()
+{
+  if [ -e ${DIR} ]; then
+    if [ ! -z "$(ls -A ${DIR})" ]; then
```
Ichimonji10: Please quote `${DIR}`. This will break if the script is called as `ostree/gen-fixtures.sh "fixtures/ostree/small tree"`.

ostree/gen-fixtures.sh
```diff
@@ -0,0 +1,80 @@
...
+      echo 1>&2 "${DIR} must be empty."
+      echo 1>&2 ""
+      exit 1
+    fi
+  else
+    mkdir -p "${DIR}"
+  fi
+}
+
+gen_tree()
```
Ichimonji10: I'm wondering what the purpose of this `gen_tree()` function is. I know that it (hopefully) copies this script into `/tmp/tree.XXXXX`, but I don't understand what purpose it serves. I can comment out the call to this function, and the end result seems the same.

jortel: This ensures that the (directory) tree contains at least 1 file.  IIRC, older versions of ostree would fail if there was not at least one file.  Just uses the script itself because it's reliable.  I also noticed that if the script exists because one of the two checks fails, the /tmp/tree.XX was not being cleaned up.  I moved setting `TREE` to `gen_tree()` as well.

Ichimonji10: > IRC, older versions of ostree would fail if there was not at least one file.

That explains a lot. Thank you.

I would recommend avoiding references to `$0`, and doing something else, like calling `touch "${TREE}/foo"`. There's [issues](http://mywiki.wooledge.org/BashFAQ/028) with `$0`.

> I also noticed that if the script exists because one of the two checks fails, the /tmp/tree.XX was not being cleaned up.

The following idiom is used in several of the pulp-fixtures scripts to ensure clean-up occurs in most situations:

```bash
# Create a workspace, and schedule it for deletion.
cleanup() { if [ -n "${working_dir:-}" ]; then rm -rf "${working_dir}"; fi }
trap cleanup EXIT  # bash pseudo signal
trap 'cleanup ; trap - SIGINT ; kill -s SIGINT $$' SIGINT
trap 'cleanup ; trap - SIGTERM ; kill -s SIGTERM $$' SIGTERM
working_dir="$(mktemp --directory)"
```

This basically says "if the script finishes executing, or if the script is nicely asked to die, then call `cleanup()`. AFAIK, Bash doesn't support exceptions, but this is pretty close.

ostree/gen-fixtures.sh
```diff
@@ -0,0 +1,85 @@
...
+"
+
+show_help()
+{
+  echo "${USAGE}"
+}
+
+_mkdir()
+{
+  if [ -e ${DIR} ]; then
```
Ichimonji10: quotes

ostree/gen-fixtures.sh
```diff
@@ -0,0 +1,85 @@
...
+  ostree commit --repo="${DIR}" -b stable --add-metadata-string=version=1.1 "${TREE}"
+  ostree commit --repo="${DIR}" -b stable --add-metadata-string=version=1.2 "${TREE}"
+  ostree commit --repo="${DIR}" -b stable --add-metadata-string=version=1.3 "${TREE}"
+  # summary
+  ostree summary -u --repo="${DIR}"
+}
+
+clean()
+{
+  if [ -d ${TREE} ]; then
```
Ichimonji10: $quotes

ostree/gen-fixtures.sh
```diff
@@ -0,0 +1,85 @@
...
+
+show_help()
+{
+  echo "${USAGE}"
+}
+
+_mkdir()
+{
+  if [ -e ${DIR} ]; then
+    if [ ! -z '$(ls -A "${DIR}")' ]; then
```
Ichimonji10: Replace the single quotes with double quotes. Due to the presence of the sub-shell (`$(...)`), there's no issues with nested double-quotes.

ostree/gen-fixtures.sh
Ichimonji10: Last thing! Quote this, and I'll merge. (And thank you!)

PR #66, opened 2017-11-27
Add new fixture, fixtures/rpm-invalid-rpm

Ichimonji10:
Add a new make target, `fixtures/rpm-invalid-rpm`. This make target
generates an invalid RPM and places it into the named directory.

Fix: https://github.com/PulpQE/pulp-fixtures/issues/65


kersommoura:
@Ichimonji10, thanks for this PR.

It will be very helpful.

PR #70, opened 2017-12-13
Drop fixtures/rpm-erratum

Ichimonji10:
This fixture is no longer being used by Pulp Smash, as per
https://github.com/PulpQE/pulp-smash/commit/efaffb64b53867e276320c723a6cca315d3f4413

Fix: https://github.com/PulpQE/pulp-fixtures/issues/68


Ichimonji10:
Related to https://github.com/PulpQE/pulp-smash/pull/825

PR #72, opened 2018-01-18
Improve array sorting code; require bash 4.4+

Ichimonji10:
Re-write the array sorting code in `rpm/gen-fixtures-delta.sh`. The new
code is better in two ways:

* It's more robust. Filenames can contain virtually any character. See:
  * http://git.savannah.gnu.org/cgit/bash.git/plain/NEWS (See the new
    features in bash 4.4, section 1 d.)
  * https://transnum.blogspot.com/2008/11/bashs-read-built-in-supports-0-as.html
  * https://stackoverflow.com/a/7442583 (See the comments.)
* It doesn't permanently alter $IFS. See:
  * https://github.com/koalaman/shellcheck/wiki/SC2207
  * https://stackoverflow.com/a/11789688

This change fixes the shellcheck warning mentioned above.

The new code takes advantage of features added in bash 4.4. As a result,
fixtures will now need to be generated on Fedora 26 or newer.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -20,11 +20,10 @@ For exact usage instructions, clone this repository and run ``make help``.
 Dependencies
 ------------
 
-The make targets and fixture generation scripts do little more than call out to
-system executables and mangle the results. It's the user's responsibility to
-ensure the necessary executables are available and usable. Dependencies are
-listed below, according to make target. Common system executables like ``fmt``,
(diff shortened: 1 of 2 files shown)


Ichimonji10:
Depends on https://github.com/pulp/pulp-ci/pull/472

Issue #73, opened 2018-02-19
Create a fixture with at least 10 iso files


kersommoura:
In order to test Pulp3 feature to delete any repo version, it was suggested to use at least 10 .iso files.

Pulp-fixtures currently generates a file dir with 3 .isos.

See: [Smash 871](https://github.com/PulpQE/pulp-smash/issues/871)

PR #74, opened 2018-02-21
Add fixtures to generate a large number of files.

kersommoura:
Re-use most of tooling already in place to generate a fixutes with a
large number of files.

This feature will be mostly used for Pulp3 tests.

Closes: #73


Ichimonji10:
The `file/gen-fixtures.sh` script is a bit antiquated anyway. I'll update it.

PR #75, opened 2018-02-23
Modernize file generation script

Ichimonji10:
Do the following:

* Make fixture creation more atomic.
* Use getopt for argument parsing, along with the `check_getopt`
  function from `rpm/common.sh`.
* Add an optional `--number` argument to the script, which controls how
  many files should be generated.
* Generate much smaller files. There's no special reason to generate
  files several megabytes in size.

### DIFF
--- a/file/gen-fixtures.sh
+++ b/file/gen-fixtures.sh
@@ -5,44 +5,79 @@
 #
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


Ichimonji10:
@kersommoura Please review this, and make #74 use it.

PR #78, opened 2018-02-28
Introducing rich dependency content

dparalen:
...in order to test rich dependencies support in pulp_rpm

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,12 @@ help:
 	@echo "    fixtures/rpm-pkglists-updateinfo"
 	@echo "        Create RPM fixtures with multiple pkglists and collections"
 	@echo "        in updateinfo.xml."
+	@echo "    fixtures/rpm-richnweak-deps"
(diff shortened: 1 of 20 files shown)


dparalen:
This is sort of WIP; need to figure out where to actually store the spec  files and/or how to expose those properly

Ichimonji10:
Feel free to make a new directory, like `rpm-rich-dependencies`, and place all new files in there. Machinery for compiling these spec files into RPMs and an RPM repository must be added. Please use mock for building RPMs. (We're also going to use mock for any new RPM fixtures that are added, and we're going to revamp our existing RPM fixtures to use mock.) For an example of how to compile RPMs into SRPMs and RPMs, see [this toy code](https://github.com/Ichimonji10/impedimenta/tree/master/bash/rpms).

dparalen:
@Ichimonji10 thanks for the input!
I'm pushing an update, please, lemme know wdyt...

Cheers,
milan

Issue #80, opened 2018-06-01
Add a repository with a very large number of files


kersommoura:
Pulp3 handles pagination. Right now, the default value of items per page is 100.

See: [Per Page](https://github.com/pulp/pulp/blob/ffd2fc78cb92ec044b170aab6b44444536db326f/pulpcore/pulpcore/app/settings.py#L112)

Create a fixture with at least 250 files to test the pagination.

See: https://github.com/PulpQE/pulp-smash/issues/918

PR #81, opened 2018-06-14
Add fixture to generate many files.

kersommoura:
Re-use most of tooling already in place to generate a fixutes with a
massive number of files.

Pulp 3 handles pagination. Right now, the default value of items per
page is 100.

This fixture will create 250 files, this one can be used to test the
pagination for Pulp 3.

Closes:#80


kersommoura:
@Ichimonji10, done.

PR #84, opened 2018-07-16
Creating an RPM with modules fixture

daviddavis:


daviddavis:
## How to test

```
cd pulp-fixtures
make fixtures/rpm-with-modules
```

Then on a Fedora 28 box with modularity setup, add this repository to /etc/yum.repos.d/zoo.repo:

```
[zoo]
name = zoo
baseurl = file:///home/vagrant/devel/pulp-fixtures/fixtures/rpm-with-modules/
```

Now you can do:

```
dnf module list
```

And so on.

PR #86, opened 2018-07-19
Package names must be in nevra format

daviddavis:


rpm/assets/modules.yaml
```diff
@@ -16,7 +16,7 @@ data:
         - walrus
   artifacts:
     rpms:
-      - walrus-5.21-1.noarch
+      - walrus-0:5.21-1.noarch
```
Ichimonji10: Isn't an exclamation point supposed to separate the epoch from the remainder of the version string?

daviddavis: From googling, it looks like it's suppose to be `name-epoch:version-release.architecture` which repoquery also seems to confirm:

```
$ repoquery --nevra '*vim*'
vim-enhanced-2:7.4.160-4.el7.x86_64
```

Ichimonji10: TIL that there are multiple ways to separate an epoch from the rest of the version. See [PEP 440 - Version epochs](https://www.python.org/dev/peps/pep-0440/#version-epochs).

daviddavis: Ah interesting.

PR #87, opened 2018-07-19
Adding in some extra modules.yaml fields

daviddavis:
These fields look like they technically aren't required by the modulemd
spec but we don't support modulemds without them so adding them in.

### DIFF
--- a/rpm/assets/modules.yaml
+++ b/rpm/assets/modules.yaml
@@ -4,6 +4,9 @@ version: 2
 data:
   name: walrus
   stream: 5.21
+  version: 20180704144203
+  context: deadbeef
+  arch: x86_64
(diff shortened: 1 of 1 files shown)


Ichimonji10:
If these fields aren't required by the modulemd spec, wouldn't it be better to make Pulp support the modulemd spec?

daviddavis:
Good question. I got some clarification around this. The short answer is that these fields are mandatory and modules without the field are invalid. The longer answer is the spec allows these fields to be optional but once the module/repo is built and the modules.yaml becomes repodata, these fields are required and must be filled in. I also tested this out and Pulp seems to fail gracefully in that it raises an error that the field is missing.

Issue #95, opened 2018-08-21
Fix syntax on `Requires` field


kersommoura:
https://github.com/PulpQE/pulp-fixtures/blob/82a010bdd985637fdaccb1fe739fedf3fa49b096/rpm-richnweak-deps/assets-specs/Cobbler.spec#L10

Verify `Requires` syntax on all the spec files. Right now, strong dependencies can not be satisfied with the current one.

When using `dnf` to install Cobbler.

```bash
  - nothing provides (scotch = 8.10 and contireau = 2.10 and tablespoon-sugar = 1.0) needed by Cobbler-1-0.noarch
```

Besides that there is no RPM called scotch. Verify name.

PR #96, opened 2018-08-21
Modify the RPM specs for rich and weak dependencies.

kersommoura:
Fix what each RPM spec provides and fix names for specs.

Using this new specs. dnf was able to install `Cobbler` an
`PanAmerican`, and its dependencies as well.

Fix:#95


kersommoura:
@dparalen, I think this PR gets us closer what we want. Please, take a look as well.

PR #102, opened 2018-09-17
Add fixture for updateinfo.xml with references.

goosemania:
Add fixture for updateinfo.xml with references


goosemania:
It won't hurt to have these references in the base updateinfo.xml in the assets.
But it will require to regenerate other updateinfo patches, so they apply cleanly.
So for now, it's a separate fixture repo.

kersommoura:
Please, can you update the README and add a simple commit message?

PR #107, opened 2018-11-06
Update README to add rpm-with-sha-512

kersommoura:
Update README to add rpm-with-sha-512. Reflect this new change.


README.rst
```diff
@@ -180,7 +183,7 @@ See ``make help``.
 
 ``gnupghome``
     The ``gpg`` executable must be available.
-
+i
```
rochacbruno: is this `i` intentional?

kersommoura: Not really. Thanks.

PR #109, opened 2018-11-06
Adding `sha 512` checksum logic for generating fixtures.

ragabala:
A user should be able to sync rpm packages that has sha 512 checksums in
the pulp repo. The current pulp repo is checksumed with only one type
`sha 256`. This commit will add logic for changing the checksum type of
the created repo by passing a parameter `--checksum-type` in the
`gen-fixtures.sh`

refer https://pulp.plan.io/issues/4007


kersommoura:
@rochacbruno, any comments?

rochacbruno:
@kersommoura Ok for me, looks like that was not a problem.

PR #110, opened 2018-11-10
Add module dependencies

goosemania:
It's needed for testing applicability feature and simple
dep solving cases.

### DIFF
--- a/rpm/assets/modules.yaml
+++ b/rpm/assets/modules.yaml
@@ -59,6 +59,17 @@ data:
   license:
     module:
       - MIT
+  dependencies:
+  - buildrequires:
+      walrus: [0.71, 5.21]
+      platform: [f28]
+    requires:
+      walrus: [0.71, 5.21]
+      platform: [f28]
+  - buildrequires:
+      platform: [el8]
+    requires:
+      platform: [el8]
   profiles:
     default:
       rpms:
@@ -82,6 +93,13 @@ data:
   license:
     module:
       - MIT
+  dependencies:
+  - buildrequires:
+      walrus: []
+      platform: [-f29, -f30]
+    requires:
+      walrus: []
+      platform: [-f29, -f30]
   profiles:
     default:
       rpms:
@@ -105,6 +123,13 @@ data:
   license:
     module:
       - MIT
+  dependencies:
+  - buildrequires:
+      walrus: [-5.21]
+      platform: [f28]
+    requires:
+      walrus: [-5.21]
+      platform: [f28]
   profiles:
     default:
       rpms:


goosemania:
I'm still verifying the meaning of those dependencies. I'll leave comments on the PR how I understand it at the moment. Specs with comments can be found here: https://github.com/fedora-modularity/libmodulemd/blob/master/spec.v2.yaml#L97-L135

rpm/assets/modules.yaml
```diff
@@ -59,6 +59,17 @@ data:
   license:
     module:
       - MIT
+  dependencies:
+  - buildrequires:
+      walrus: [0.71, 5.21]
+      platform: [f28]
```
goosemania: For F28, `walrus` module is required, either stream 0.71 or 5.21

rpm/assets/modules.yaml
```diff
@@ -59,6 +59,17 @@ data:
...
+    requires:
+      walrus: [0.71, 5.21]
+      platform: [f28]
```
goosemania: For F28, `walrus` module is required, either stream `0.71` or `5.21`

rpm/assets/modules.yaml
```diff
@@ -59,6 +59,17 @@ data:
...
+  - buildrequires:
+      platform: [el8]
+    requires:
+      platform: [el8]
```
goosemania: For EL8, no additional modules are required. (but this module can run on EL8?)

rpm/assets/modules.yaml
```diff
@@ -82,6 +93,13 @@ data:
   license:
     module:
       - MIT
+  dependencies:
+  - buildrequires:
+      walrus: []
+      platform: [-f29, -f30]
+    requires:
+      walrus: []
+      platform: [-f29, -f30]
```
goosemania: This module requires any platform except F29 or F30 and it requires `walrus` module of any stream.

rpm/assets/modules.yaml
```diff
@@ -105,6 +123,13 @@ data:
   license:
     module:
       - MIT
+  dependencies:
+  - buildrequires:
+      walrus: [-5.21]
+      platform: [f28]
+    requires:
+      walrus: [-5.21]
+      platform: [f28]
```
goosemania: This modules requires platform F28 and `walrus` module of any version except `5.21`

goosemania:
Here are docs with examples for dependencies https://docs.fedoraproject.org/en-US/modularity/making-modules/defining-modules/

It seems like my understanding is correct.

kersommoura:
@goosemania, mind adding a simple commit message for future reference?why this change was needed.

kersommoura:
@rochacbruno, any other comment?

rochacbruno:
@kersommoura LGTM

PR #111, opened 2018-11-19
Add rpm-with-modules for el8

parthaa:


kersommoura:
@parthaa, thanks for your PR. Please, Can you add a simple commit message for future reference?

PR #114, opened 2019-01-15
Update .travis.yml

kersommoura:
Update .travis.yml to install shellcheck using package manager. There
was a problem and the shellcheck version was pinned before.
The problem has been solved, and shellcheck is currently in the version
`0.6.0`.

Besides that, update new error found using the new version of
shellcheck.

See: https://github.com/koalaman/shellcheck/releases/tag/v0.6.0

### DIFF
--- a/ostree/gen-fixtures.sh
+++ b/ostree/gen-fixtures.sh
@@ -29,7 +29,7 @@ show_help()
 _mkdir()
 {
   if [ -e "${DIR}" ]; then
-    if [ ! -z "$(ls -A "${DIR}")" ]; then
+    if [ -n "$(ls -A "${DIR}")" ]; then
       echo 1>&2 ""
       echo 1>&2 "${DIR} must be empty."
       echo 1>&2 ""
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,13 +11,7 @@ before_install:
     - pip install --upgrade pip
     - pip install pylint
     - deactivate
-    - cabal update
-    - git clone https://github.com/koalaman/shellcheck.git
-    - cd shellcheck
-    - git checkout v0.4.4
-    - cabal install
-    - cd ..
-    - export PATH="$HOME/.cabal/bin:$PATH"
+    - sudo apt-get install -y shellcheck
     - which shellcheck
     - source ~/env/bin/activate
 script:


ostree/gen-fixtures.sh
```diff
@@ -29,7 +29,7 @@ show_help()
 _mkdir()
 {
   if [ -e "${DIR}" ]; then
-    if [ ! -z "$(ls -A "${DIR}")" ]; then
+    if [ -n "$(ls -A "${DIR}")" ]; then
```
bherrin3: I believe you need to negate this, or you are going to invert the logic here seeing if the DIR is empty.

Meaning, if there is a `$DIR` AND `$DIR` has contents, it is going to state "`$DIR` is empty." and `exit 1`. 

I don't think that is the intent here? 

Simply negating the expression should fix that, unless I am missing some new intended logic?

```suggestion
    if [ ! -n "$(ls -A "${DIR}")" ]; then
```
By running this as an example with a test DIR `test` and touching a file in that dir, `test/chickens`, the following execution resulted with the exit 1 (assume failed RC when non-zero... kinda):
```
[herring@redherring 114]$ cat test.sh 
#!/bin/bash

DIR=test

  if [ -e "${DIR}" ]; then
    if [ -n "$(ls -A "${DIR}")" ]; then
      echo 1>&2 ""
      echo 1>&2 "${DIR} must be empty."
      echo 1>&2 ""
      exit 1
    fi
  else
    mkdir -p "${DIR}"
  fi
[herring@redherring 114]$ ll
total 8
drwxrwxr-x. 2 herring herring 4096 Jan 15 09:22 test
-rwxrwx---. 1 herring herring  222 Jan 15 09:22 test.sh
[herring@redherring 114]$ ll test
total 0
-rw-rw-r--. 1 herring herring 0 Jan 15 09:22 chickens
[herring@redherring 114]$ ./test.sh 

test must be empty.

[herring@redherring 114]$ echo $?
1
```

kersommoura:
``` text
-z "$str" True if length of string is zero
-n "$str True if length of string is non-zero
```
If we adopt your suggestion the linter is not happy.

```bash
    if [ ! -n "$(ls -A "${DIR}")" ]; then
         ^-- SC2236: Use -z instead of ! -n.
```

We can just ignore this case and keep the logic as it was before.

PR #123, opened 2019-07-03
Add new fixture with errata pkglist

kersommoura:
Add new fixture with errata name already used by other repository, but
with different pkglist.

See: https://pulp.plan.io/issues/4952


bherrin3:
Re-built the fixtures on Upshift -- all built without issue.

Assuming the integrity of the fixture from the test in `Pulp-2-Tests`

PR #125, opened 2019-07-23
Add missing file fixture

daviddavis:


daviddavis:
Looks like this already exists ("file-mixed").

PR #126, opened 2019-08-05
Add script, create and push to dockerhub

asmacdo:


kersommoura:
@asmacdo, thank you. This is really helpful. As we talked perhaps we could add a note to the commit message since it is not the goal of those scripts to be generated together with the *make all*.

asmacdo:
Updated to reflect that this directory is not separate from the rest of the fixtures. For that reason, I am not deliberately leaving it out of the top level README.

PR #127, opened 2019-12-10
Add fixtures needed to test BZ #1741011

rochacbruno:
An updateinfo with a `version="<str>"` should be valid
and this fixture to test it.

see: https://bugzilla.redhat.com/show_bug.cgi?id=1741011


rochacbruno:
I really need this fixture merged

PR #134, opened 2020-01-21
Tweaks to Dockerfile

daviddavis:


daviddavis:
This change is more difficult than I realized since files cannot be moved inside a container. I've opened an issue to investigate possible solutions and look for other optimizations. 

https://pulp.plan.io/issues/6039

PR #136, opened 2020-01-27
Use multi stage dockerfile

mdellweg:


mdellweg:
@daviddavis I believe this is not clean enough, but i wanted to show you how i think a multi stage build can help to generate that container.

daviddavis:
This looks great. Why do you say it's not clean enough?

mdellweg:
I only copied the apt repo fixture script over from `pulp_deb`. It didn't lint by then, also it does not yet use the Makefile.

mdellweg:
Also for just serving the stuff, there is probably a better container with nginx preinstalled available.

daviddavis:
If you want to do the steps you outlined, I'd be happy to merge this. Also, I think we could probably use [the nginx image from docker hub](https://hub.docker.com/_/nginx).

mdellweg:
Is using this container properly tested? I am not entirely sure, the second copy instruction does not overwrite the result of the first. But i marked the PR as ready, as i fixed the linting. And using the nginx container should be another PR.

mdellweg:
My experiment with buildah shows, that the directories are merged. :heavy_check_mark:

daviddavis:
I tested this out locally but I'm seeing the fixtures now in /usr/share/nginx/html/ instead of /usr/share/nginx/html/fixtures/. Are you seeing the same thing?

daviddavis:
I added a small test that might help you to verify the container:

https://github.com/pulp/pulp-fixtures/pull/141

mdellweg:
> I tested this out locally but I'm seeing the fixtures now in /usr/share/nginx/html/ instead of /usr/share/nginx/html/fixtures/. Are you seeing the same thing?

Haven't checked that...

PR #138, opened 2020-02-04
Adding RPM manual fixtures

fabricio-aguiar:
- [rpm-kickstart](https://repos.fedorapeople.org/pulp/pulp/fixtures/rpm-kickstart/)
- [rpm-with-non-utf-8](https://repos.fedorapeople.org/pulp/pulp/fixtures/rpm-with-non-utf-8/)


Makefile
```diff
@@ -152,6 +152,7 @@ fixtures: fixtures/docker \
 	fixtures/file-perf \
 	fixtures/file-mixed \
 	fixtures/python-pypi \
+	fixtures/rpm-manually-created-fixtures \
```
daviddavis: I would just create a new fixture called "rpm-kickstart" and uncomment/update "rpm-with-non-utf-8". I imagine that someone might want to just generate one or the other but not both.

PR #143, opened 2020-02-06
Sign debian fixture repo

mdellweg:


mdellweg:
ping @daviddavis

daviddavis:
Sorry for the delay. :)

mdellweg:
Thank you! @daviddavis
I like the fact, that it was sooo easy to reuse the gpghome dir in the other container.

PR #145, opened 2020-02-12
Fix copying of debian fixtures

mdellweg:


.travis.yml
```diff
@@ -19,6 +19,7 @@ jobs:
         - docker build -t quay.io/pulp/pulp-fixtures .
         - docker run -d -p 8000:80 quay.io/pulp/pulp-fixtures
         - curl --fail localhost:8000/fixtures/file/PULP_MANIFEST
+        - curl --fail localhost:8000/fixtures/debian/dists/ragnarok/Release
```
quba42: Adds a test to ensure there is at least some content from each build container present. ACK!

debian/gen-fixtures.sh
```diff
@@ -51,5 +51,6 @@ reprepro -C jotunheimr includedeb ragnarok jotunheimr/*.deb
 
 rm dists/ragnarok/jotunheimr/binary-armeb/Packages
 
-cp -r --no-preserve=mode --reflink=auto "${TMPDIR}"/dists "${OUTPUTDIR}"
-cp -r --no-preserve=mode --reflink=auto "${TMPDIR}"/pool "${OUTPUTDIR}"
+mkdir -p "${OUTPUTDIR}"
```
quba42: Ensures the two following commands actually both work in the same way. ACK!

debian/gen-fixtures.sh
```diff
@@ -51,5 +51,6 @@ reprepro -C jotunheimr includedeb ragnarok jotunheimr/*.deb
...
+cp -r --no-preserve=mode --reflink=auto "${TMPDIR}"/dists -t "${OUTPUTDIR}"
+cp -r --no-preserve=mode --reflink=auto "${TMPDIR}"/pool -t "${OUTPUTDIR}"
```
quba42: Do we need both the `-t` and the `mkdir -p`?

mdellweg: No, the `-t` would fail if the directory didn't exist. So it is there as a safeguard.

PR #146, opened 2020-02-19
Add a Debian fixture without the Suite field

quba42:
Adds the `nosuite` distribution to the Debian fixtures.
As the name implies this distribution does not contain the optional `Suite` field in its Release file.
This will form the basis for a test to ensure `pulp_deb` can synchronize such a repository.


quba42:
Since it just adds a new component/distribution/codename that is not yet being synced by the existing pulp_deb tests, it definitely won't break anything.
Amending the pulp_deb tests to use this new fixture has already revealed several bugs, so it is very useful to have.
I will remove the WIP.

quba42:
@mdellweg Could you have a quick look at this?

PR #149, opened 2020-03-04
Add .dockerignore file

quba42:
The main reason I did this is to prevent build artifacts of local invocations of `make` from influencing the container build. (The Dockerfile copies the entire repository into the build container including any local build artifacts.)

All the refactoring of the Makefile is just a simple consequence of the primary aim. Since I excluded the `fixtures` folder from being added to the container image, I needed the `make` build to be independent of the existence of this folder.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+/fixtures/
+/gnupghome/
+/.git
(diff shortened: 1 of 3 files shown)


mikedep333:
@quba42 Very cool. Now I know about .dockerignore files.

PR #150, opened 2020-03-16
Create debian_update fixture

mdellweg:
Provide a repository that represents the state of the next day.


quba42:
The resultant fixture looks good to me.
My only real question is whether the new fixtures should receive their own make target, like for the invalid fixtures.
Since they end up in their own directory this might be nice.

mdellweg:
Maybe nice, but they are not really a "new" and independent fixture, since the not updated packages must be the same (bit wise). I do not see a way to ensure this in a separate make target.

PR #152, opened 2020-03-31
Complex deb distributions

quba42:
Adds fixtures for repositories stuctured like debian-security and ubuntu-backports.
These two repositories cover the most common cases of APT repositories with a non standard distribution path, which demands test coverage.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,6 +157,7 @@ all-skipped: \
 all-debian: \
 	fixtures/debian \
 	fixtures/debian-invalid \
+	fixtures/debian-complex-dists \
 
 all-fedora: \
 	fixtures/drpm-signed \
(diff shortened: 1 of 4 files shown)


quba42:
I am getting some failed tests with these changed fixtures, so perhaps it was premature opening the PR.
I will investigate...

quba42:
I ran some more tests, and I did get some failures, but they were sporadic/not reproducible.
As a result, I think they are probably a timing or order issue that is independent of these new fixtures.

I am unsure whether it is best to merge these changes now, or wait until the tests that will use them are done. (I can't give a time frame, but we are working on them).

daviddavis:
> I am unsure whether it is best to merge these changes now, or wait until the tests that will use them are done. (I can't give a time frame, but we are working on them).

I think this is up to you. We're happy either way.

quba42:
I am merging this now, since https://github.com/pulp/pulp_deb/pull/167 now depends on it.

PR #154, opened 2020-04-29
Advisories test fixtures

pavelpicka:
new fixtures for advisory testing added with pulp/pulp_rpm#1689

re: #4142
https://pulp.plan.io/issues/4142

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,9 +50,15 @@ help:
 	@echo "        Create a dummy Puppet module."
 	@echo "    fixtures/python-pypi [base_url=...]"
 	@echo "        Create a PyPI Python repository."
+	@echo "    fixtures/rpm-advisory-incomplete-package-list"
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -50,9 +50,15 @@ help:
 	@echo "        Create a dummy Puppet module."
 	@echo "    fixtures/python-pypi [base_url=...]"
 	@echo "        Create a PyPI Python repository."
+	@echo "    fixtures/rpm-advisory-bad"
```
goosemania, line 53: Why is it bad? If the goal is to make  an incomplete advisory, maybe it makes sense to name it the same way? rpm-advisory-inomplete?

pavelpicka: It is not really 'incomplete', this advisory will be with same id, version and dates as in 'unsigned' but missing one package from collection. 
Agree to try to re-word it. Will use something like 'rpm-advisory-incomplete-package-list'.

PR #160, opened 2020-06-04
Moving /fixtures to /

daviddavis:
fixes #6880


daviddavis:
@quba42 would like your feedback on this. This would move https://fixtures.pulpproject.org/fixtures/ to https://fixtures.pulpproject.org/.

PR #167, opened 2020-06-16
Fix entrpoint bug and add base_url check

daviddavis:


.travis.yml
```diff
@@ -17,9 +17,11 @@ jobs:
       before_install: skip
       script:
         - docker build -f Containerfile -t pulp/pulp-fixtures .
-        - docker run -d -p 8000:80 pulp/pulp-fixtures
+        - docker run -d -e BASE_URL=http://example.com -p 8000:80 pulp/pulp-fixtures
+        - sleep 3
```
daviddavis, line 21: The sed replace for BASE_URL in the entrypoint script takes a couple seconds so we have to add a sleep here.

bmbouter: Would you mind putting ^ in a comment after the sleep 3 line?

PR #173, opened 2020-06-23
Support for use auth token

pavelpicka:
Removing 'sed' injections for nginx config.
Using custom nginx.conf instead.
Adding ability to use auth token (e.g. '?secret') and check for wrong tokens.
Adds two simple checks for travis. 

Now there is two ways to reach repositories.

- direct e.g. pulp-fixtures/rpm-signed/
- with auth token e.g. pulp-fixtures/rpm-signed/?secret

closes: #7004
https://pulp.plan.io/issues/7004

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -38,8 +38,8 @@ RUN rm /usr/share/nginx/html/index.html
 COPY --from=fedora-build pulp-fixtures/fixtures /usr/share/nginx/html
 COPY --from=debian-build pulp-fixtures/fixtures /usr/share/nginx/html
 
-# turn on autoindex
-RUN sed -i -e '/location \/ {/a autoindex on\;' /etc/nginx/conf.d/default.conf
+# use custom nginx.conf
+COPY common/nginx.conf /etc/nginx/nginx.conf
 
 EXPOSE 80
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,9 @@ jobs:
         - sleep 2 # it can take a couple seconds for sed to run and nginx to boot
         - curl --fail localhost:8000/file/PULP_MANIFEST
         - curl --fail localhost:8000/debian/dists/ragnarok/Release
+        - curl -L localhost:8000/rpm-unsigned/?badtoken | grep "Wrong auth token"
+        - curl --fail -L localhost:8000/rpm-unsigned/?secret
+        - curl --fail -L localhost:8000/file-large/?parameter
         - '[[ $(curl localhost:8000/rpm-mirrorlist-good) == "http://example.com/rpm-unsigned/" ]] || exit 1'
       deploy:
         - provider: script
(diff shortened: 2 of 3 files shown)


Containerfile
```diff
@@ -40,6 +40,11 @@ COPY --from=debian-build pulp-fixtures/fixtures /usr/share/nginx/html
 
 # turn on autoindex
 RUN sed -i -e '/location \/ {/a autoindex on\;' /etc/nginx/conf.d/default.conf
+# tune nginx config file for auth token
+RUN mkdir -p /pulp-fixtures/common
+ADD common /pulp-fixtures/common
+RUN sed -i -e '/http {/r /pulp-fixtures/common/nginx_auth_token_map' /etc/nginx/nginx.conf
+RUN sed -i -e '/server {/r /pulp-fixtures/common/nginx_auth_token_cond' /etc/nginx/conf.d/default.conf
```
mdellweg, line 47: Maybe, at this point we should just copy our version of nginx.conf into the container image. This sed surgery start to look adventurous to me.
Do not get me wrong. I love a good sed script. 👨‍💻 But it is also a bit dangerous.

mdellweg:
diff patch may also be worth considering.

daviddavis:
I'd probably lean towards maintaining our own copies of nginx.conf and default.conf.

daviddavis:
If I understand correctly, this adds a check for `?secret` at the end of some paths? My nginx config fu is not strong.

Also, any chance you could add a test or two here: https://github.com/pulp/pulp-fixtures/blob/0d1949ac762a325ca1cf95dbde12290a9417641e/.travis.yml#L22-L24

pavelpicka:
In my opinion is just adding 'few lines' on right place. But we can have our copy/version of nginx.conf. In that case I'd prefer to not keep nginx.conf & default.conf but join them together as one file.

Containerfile
```diff
@@ -38,8 +38,12 @@ RUN rm /usr/share/nginx/html/index.html
...
 COPY --from=debian-build pulp-fixtures/fixtures /usr/share/nginx/html
 
-# turn on autoindex
-RUN sed -i -e '/location \/ {/a autoindex on\;' /etc/nginx/conf.d/default.conf
+# tune nginx config file
+RUN mkdir -p /pulp-fixtures/common
+ADD common /pulp-fixtures/common
+RUN echo "TED SE TO DEJE!"
+RUN rm -f /etc/nginx/nginx.conf
+RUN cp /pulp-fixtures/common/nginx.conf /etc/nginx/nginx.conf
```
mdellweg, line 46: Would that not be enough?

```suggestion
COPY common/nginx.conf /etc/nginx/nginx.conf
```

daviddavis:
LGTM. @mdellweg what do you think?

daviddavis:
Actually I just had a thought: would it be possible to limit this to paths with `rpm` in the path? I worry about this interfering with other content types.

pavelpicka:
Updated that conditional check for auth token should apply only for repositories contains 'rpm' in URL. 
Added simple test that '?parameter' doesn't affect 'file' repository.

daviddavis:
Thanks @pavelpicka!

PR #175, opened 2020-07-01
Adding in ostree, puppet, and docker fixtures

daviddavis:
The puppet fixture script was disabled a long time ago. We should look
into maybe re-enabling it.

The docker asset would be hard to generate in a container because it'd
require running docker inside docker.


mdellweg:
You know there is the dind container? But the question is, what do we want to generate as container fixtures?
If this is what was found at the old address, i think we are safe to merge it as is.

daviddavis:
@mdellweg thanks I did not know there was a dind container. Here's what the docker fixture is doing right now:

https://github.com/pulp/pulp-fixtures/blob/e31a3b02296d89b02f4c309ade044a673e1e9c21/docker/gen-fixtures.sh#L41-L44

PR #181, opened 2020-07-23
Fixture for distribution tree testing

pavelpicka:
fixtures for distribution tree with packages, two addons, two variants (one external)


daviddavis:
This all looks good to me. Out of curiosity, does this kickstart actually work in anaconda? Also, I wonder if we could cut down on the size of this kickstart. Maybe we can remove the 32 bit files?

pavelpicka:
I agree, I think it is possible to make it slimmer.

daviddavis:
This is excellent. The variants seem a bit random in terms of their naming and the packages they include. Any chance you could use a logical pattern? Maybe something like a Land variant with lion, wolf, etc and a Sea variant with shark, dolphin?

PR #186, opened 2020-08-10
Distribution tree repo variants

pavelpicka:
Variants of RPM distribution tree with:
- change (added package) in main repo/variant
- change (added package) in addon
- change (added package) in variant


daviddavis:
This looks good to me. Would it be possible or easy to de-duplicate some of the code?

pavelpicka:
@daviddavis possible but not easy and imo won't save many lines

PR #187, opened 2020-08-12
Add a valid debian repo containing no packages

quba42:
Fixture needed for a test driven approach to solving https://pulp.plan.io/issues/6920

### DIFF
--- a/debian/conf/distributions
+++ b/debian/conf/distributions
@@ -10,3 +10,9 @@ Description: This is a valid Debian repository.
 Architectures: ppc64
 Components: asgard
 SignWith: 6EDF301256480B9B801EBA3D05A5E6DA269D9D98
+
+codename: ginnungagap
+Description: This repository is valid, but contains no packages!
+Architectures: armeb ppc64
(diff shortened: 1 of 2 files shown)


quba42:
It appears that adding another distribution to the existing main fixtures currently breaks the following 4 tests:
```
django-admin test pulp_deb.tests.functional.api.test_download_policies
```

I see two possible ways of dealing with this:

1. Always add new fixtures as a new fixtures repository, rather than a new distribution in an existing repository.
1. Change the tests to be explicitly tied to the existing distributions.

I currently prefer (2), since it will cause less of a mess in the fixtures repo.
@mdellweg Any thoughts?

quba42:
I ran the pulp_deb tests against a locally built container with this change.
I had four failed tests, however those also failed with a locally built container without this change, so it is probably something to do with my environment and not with the change (all tests have been passing on recent pulp_deb PRs).

The failed tests were all in:
```
django-admin test pulp_deb.tests.functional.api.test_download_policies
```
I am nevertheless removing the "Draft" status.

PR #190, opened 2020-08-27
Added new X-PyPI-Last-Serial header to python fixture pages

gerrod3:
The python plugin expects this header to be there to perform a cache check to see if it needs to update a package. Since the fixtures are always used when testing from a clean state, setting this to a high constant will make the plugin always update the package.

[noissue]

### DIFF
--- a/common/nginx.conf
+++ b/common/nginx.conf
@@ -48,6 +48,9 @@ http {
             autoindex on;
             root   /usr/share/nginx/html;
             index  index.html index.htm index.json;
+            location ~* ^/python-pypi/.*\.(html|json)$ {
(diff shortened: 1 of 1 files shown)


gerrod3:
Before merging I think I broke the redirects for links without a trailing backslash, at least for locally hosted fixtures.  The links in the simple directory don't have a trailing backslash and nginx redirects to the same url without the port number causing a 404.  I am not sure how to fix this redirect issue.

gerrod3:
Ok I think this PR should be good to go.

dralley (approved):

dralley:
@dkliban If this looks good to you, could you merge?

mdellweg (approved):
I can.

PR #191, opened 2020-08-28
Prevent nginx from stripping port when it redirects

daviddavis:
Nginx drops the port number when it redirects ('localhost:8000/fixtures/python-pypi/simple/package' -> 'localhost/fixtures/python-pypi/simple/package/')


daviddavis:
Fixed in https://github.com/pulp/pulp-fixtures/pull/194

PR #193, opened 2020-08-29
Remove the old kickstart fixture

daviddavis:
ref #7040


daviddavis:
This is waiting on https://github.com/pulp/pulp_rpm/pull/1832

daviddavis:
@fao89 @goosemania @pavelpicka I think this can be merged now. Can you confirm?

daviddavis:
Actually, I think merging this will cause older release branches to fail (e.g. pulp_rpm 3.5, 3.6, etc).

pavelpicka:
To avoid fails of old release, can we leave symlink from old repo to new one?

daviddavis:
I don't think a symlink would work for two reasons: (a) syncing the new fixture will cause content to get stuck in the database because of [the kickstart orphan cleanup bug](https://pulp.plan.io/issues/7440) and (b) the tests are written using this old fixture (eg https://git.io/JUa4H).

I think if we want to remove this fixture, we need to either (1) backport the kickstart orphan cleanup bug and test update to all supported releases, or (2) wait until 3.6 is no longer supported and we no longer are doing releases from < 3.7. I'm fine with either option but if we go with (2) I'd want to add a deprecation notice somewhere to the rpm-kickstart fixture.

fao89:
> I don't think a symlink would work for two reasons: (a) syncing the new fixture will cause content to get stuck in the database because of [the kickstart orphan cleanup bug](https://pulp.plan.io/issues/7440) and (b) the tests are checking the distribution tree structure (eg https://git.io/JUa4H).
> 
> I think if we want to remove this fixture, we need to either (1) backport the kickstart orphan cleanup bug and test update to all supported releases, or (2) wait until 3.6 is no longer supported and we no longer are doing releases from < 3.7.
> 
> I'm fine with either solution but if we go with (2) I'd want to add a deprecation notice somewhere to the rpm-kickstart fixture.

I would go with (2)

pavelpicka:
I like option 2 too. So maybe instead this PR just add the comment line to remove when 3.6 got retired.

goosemania:
+1 to wait until 3.8 is out

daviddavis:
Cool, I added a deprecation notice: https://github.com/pulp/pulp-fixtures/pull/198

daviddavis:
@pulp/rpm can this be merged? It looks like [3.6](https://github.com/pulp/pulp_rpm/blob/3.6/pulp_rpm/tests/functional/constants.py#L331) was the last branch to use this fixture and we haven't done a z-stream of it 3.6 since November, 2020.

goosemania (approved):

PR #197, opened 2020-10-08
Repository metadata fixture

pavelpicka:
adding rpm-repo-metadata with 'productid' type metadata


daviddavis:
From a pulp-fixtures point of view, this looks good. I do not have enough RPM knowledge to evaluate this though. Perhaps @goosemania, you could also review?

pavelpicka:
we don't have this content type in any rpm fixture, so adding it to better tests.

goosemania (approved):

PR #199, opened 2020-10-15
Add Debian flat repository format fixtures

ThikaXer:
fixes #7502
https://pulp.plan.io/issues/7502

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,12 @@ help:
 	@echo "        Create Debian repository fixtures."
 	@echo "    fixtures/debian-invalid"
 	@echo "        Create invalid Debian repository fixtures."
+	@echo "    fixtures/debian-complex-dists"
+	@echo "        Create Debian repository fixtures with complex distribution."
+	@echo "    fixtures/debian-missing-architecture"
+	@echo "        Create Debian repository fixtures with missing Packages files for some architectures."
+	@echo "    fixtures/debian-flat"
+	@echo "        Create Debian repository fixtures in flat repository format."
 	@echo "    fixtures/docker"
 	@echo "        Create Docker fixture data."
 	@echo "    fixtures/drpm-signed"
@@ -171,6 +177,7 @@ all-debian: \
 	fixtures/debian-invalid \
 	fixtures/debian-complex-dists \
 	fixtures/debian-missing-architecture \
+	fixtures/debian-flat \
 
 all-fedora: \
 	fixtures/drpm-signed \
@@ -251,6 +258,9 @@ fixtures/debian-complex-dists: fixtures gnupghome
 
 fixtures/debian-missing-architecture: fixtures gnupghome
 	GNUPGHOME=$$(realpath -e gnupghome) debian/gen-missing-architecture-fixtures.sh $@
+
+fixtures/debian-flat: fixtures
+	debian/gen-flat-repo-fixtures.sh $@
 
 fixtures/docker: fixtures
 	docker/gen-fixtures.sh $@
--- a/debian/gen-flat-repo-fixtures.sh
+++ b/debian/gen-flat-repo-fixtures.sh
@@ -1,0 +1,32 @@
+#!/bin/sh
+
+set -eux
+
+SRCDIR=$(readlink -f "$(dirname "$0")")
+TMPDIR=$(mktemp -d)
+OUTPUTDIR=$(realpath "${SRCDIR}/../$1")
+
+trap 'rm -rf "${TMPDIR}"' EXIT
+
+cd "${TMPDIR}"
+(
+  for CTL in "${SRCDIR}"/asgard/*.ctl
+  do
+    equivs-build --arch ppc64 "${CTL}"
+  done
+)
+
+dpkg-scanpackages . | tee ./Packages | gzip > ./Packages.gz
+
+echo 'Suite: mythology
+Codename: ragnarok
+Architectures: ppc64
+Components: asgard' > Release
+
+apt-ftparchive release . >> Release
+
+mkdir -p "${OUTPUTDIR}"/nest/fjalar/
+cp -r --no-preserve=mode --reflink=auto "${TMPDIR}"/. -t "${OUTPUTDIR}"
+cp -r --no-preserve=mode --reflink=auto "${TMPDIR}"/Release -t "${OUTPUTDIR}"/nest/fjalar/
+cp -r --no-preserve=mode --reflink=auto "${TMPDIR}"/Packages -t "${OUTPUTDIR}"/nest/fjalar/
+cp -r --no-preserve=mode --reflink=auto "${TMPDIR}"/Packages.gz -t "${OUTPUTDIR}"/nest/fjalar/
--- a/Containerfile
+++ b/Containerfile
@@ -21,7 +21,12 @@ RUN make -C pulp-fixtures all-fedora base_url=http://BASE_URL
 # === Build fixtures (Debian) =================================================
 FROM debian:stretch AS debian-build
 
-RUN apt-get update && apt-get -y install equivs reprepro
+RUN apt-get update
+RUN apt-get -y install \
+              equivs \
+              reprepro \
+              dpkg-dev \
+              apt-utils
 
 RUN mkdir -p /pulp-fixtures/fixtures
 ADD Makefile /pulp-fixtures


debian/gen-flat-repo-fixtures.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+
+set -eux
+
+sudo apt-get install dpkg-dev apache2 dpkg-sig
```
mdellweg, line 5: This is run in the (transient) container as root.

```suggestion
apt-get install dpkg-dev apache2 dpkg-sig
```

ThikaXer: Thank you for your comment! Changes have been integrated.

debian/gen-flat-repo-fixtures.sh
```diff
@@ -0,0 +1,37 @@
...
+  done
+)
+
+curl -L https://github.com/pulp/pulp-fixtures/raw/master/common/GPG-PRIVATE-KEY-pulp-qe | gpg --import
+echo "6EDF301256480B9B801EBA3D05A5E6DA269D9D98:6:" | gpg --import-ownertrust
+
+dpkg-sig --sign builder ./*.deb
+
+apt-ftparchive packages ../debian-flat > Packages
+sed -i 's!../debian-flat!debian-flat!' ./Packages
```
mdellweg, line 26: Why is this manipulation needed? Can't you just `cd ../debian-flat` and call `apt-ftparchive` there?

BTW: I thought a flat repo was created with `dpkg-scanpackages | gz > Packages.gz` but that might be outdated information.

debian/gen-flat-repo-fixtures.sh
```diff
@@ -0,0 +1,37 @@
...
+gzip -c Packages > Packages.gz
+
+echo 'Suite: mythology
+Codename: ragnarok-flat
```
mdellweg, line 30: Not sure i like the "-flat" part of that name.
Is it needed to not collide with other tests?

mdellweg (commented):
Thank you for adding those here.
I left a few questions for things that look suspicious to me.

debian/gen-flat-repo-fixtures.sh
ThikaXer: The problem is, that the Packages file still has the values for path starting with "./" which I need to replace with "debian-flat/", otherwise pulp cannot find the packages. One could also go inside the folder, create the Packages file, change the entries and go outside again...
I think one could also use dpkg-scanpackages, but as I understand the manpage, that only creates a .gz or .xz file. Please see https://pulp.plan.io/issues/7768?issue_count=26&issue_position=1&next_issue_id=7756 as a related issue. I had troubles synchronising repositories in pulp_deb which have only a Packages.gz or Packages.xz file.

debian/gen-flat-repo-fixtures.sh
ThikaXer: No, they are not necessary. I deleted the "-flat" for Codename and Components now.

debian/gen-flat-repo-fixtures.sh
mdellweg: What i am thinking here is, "If the flat-repo-format is a _thing_, there must be tools around to create it. And those tools would not require you to make such strange manipulations". So i urge you to find those tools (no, i don't mean pulp XD).

On the other hand changing this fixture in a way that pulp can sync it is really the wrong direction.

The fixture should be reflecting a format that is valid and consumable by apt. And once you have that you try to make pulp understand it. And as far as i remember, the easiest way to create a repo is:
Throw some packages in a directory and call `dpkg-scanpackages` (It might depend on where you call it how it references the files relatively.)

Maybe the one including a `Release` file is already a different format.

https://linux.die.net/man/1/dpkg-scanpackages

ThikaXer: Thanks for your ideas! Still, there are some issues for me that I cannot quite see through:

1) I don't really think that flat repo format is "the big thing". At least, it's not recommended as "best practice". Still, some people use it to build packages manually, at least I could not find any app that creates packages in flat repo format. If someone knows how to use an app, they would use an app that creates a nice repo ie not flat, wouldn't they? Still, I changed the generation as I would do it manually. What do you think? Not really satisfying, is it?
Why I need to change the path for pulp, I don't really get either. For the bash commands, I could insert as many "./" as I want in a path and it still can find the correct path. However, Pulp cannot deal with "./" in a path, or rather the os.path python module then maybe returns wrong file paths or components... But maybe I can change that in the synchronizing.py somehow...

2) Do the SHA and MD5 values depend on which folder you are in when building the Packages file? When I tried it (once from outside, once from inside the folder), the SHA and MD5 did not change...

2) The flat repos I've found in the Internet so far all have Release files, so I think we can expect that. Also the debian repo standard requires a Release file. What do you think?

Thanks for your help!

Makefile
```diff
@@ -20,6 +20,12 @@ help:
 	@echo "        Create Debian repository fixtures."
 	@echo "    fixtures/debian-invalid"
 	@echo "        Create invalid Debian repository fixtures."
+	@echo "    fixtures/debian-complex-dists"
+	@echo "        Create Debian repository fixtures with complex distribution."
+	@echo "    fixtures/debian-missing-architecture"
+	@echo "        Create Debian repository fixtures with missing Packages files for some architectures."
```
mdellweg, line 26: :+1:

debian/gen-flat-repo-fixtures.sh
```diff
@@ -0,0 +1,39 @@
...
+  for CTL in "${SRCDIR}"/asgard/*.ctl
+  do
+    equivs-build --arch ppc64 "${CTL}"
+  done
+)
+
+curl -L https://github.com/pulp/pulp-fixtures/raw/master/common/GPG-PRIVATE-KEY-pulp-qe | gpg --import
+echo "6EDF301256480B9B801EBA3D05A5E6DA269D9D98:6:" | gpg --import-ownertrust
+
+dpkg-sig --sign builder ./*.deb
```
mdellweg, line 23: This signs the packages. Right? We only do signed Metadata in the other debian fixtures. And i doubt the plugin needs this extra time consuming step.

debian/gen-flat-repo-fixtures.sh
```diff
@@ -0,0 +1,39 @@
...
+
+cd ..
+dpkg-scanpackages debian-flat/ | gzip > Packages.gz
+mv Packages.gz debian-flat/Packages.gz
+cd debian-flat
+gzip -cd Packages.gz > Packages
```
mdellweg, line 29:
```suggestion
dpkg-scanpackages debian-flat/ | tee debian-flat/Packages | gzip > debian-flat/Packages.gz
cd debian-flat
```

I'm still not really sure about the cd dance here, but let's make the Packages files in one go at least.

mdellweg (commented):
I get, that we both don't know enough about the actual "format" of flat repositories.
I left a few comments to make the build more efficient, but then we should be good to go.
If we matched what is out there in the wild, only time will tell.

debian/gen-flat-repo-fixtures.sh
ThikaXer: Done. Thanks :)

debian/gen-flat-repo-fixtures.sh
ThikaXer: Done. Thanks :)

mdellweg:
@quba42 I want your eyes on this. If you are sufficiently convinced, this is mimicking what is out there in the wild (I still can't believe there's not tool for it.) this is good to go.

mdellweg:
Oh, one last thing: can we maybe reference that it's about a debian repository in the commit message header?

debian/gen-flat-repo-fixtures.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
+
+set -eux
+
+yes | apt-get install dpkg-dev apt-utils
```
quba42, line 5: These build dependencies should go in the `debian-build` container in the `Containerfile`.
Ideally this script should not require root to run locally. :wink:

I will add a suggestion commit to show exactly what I mean.

ThikaXer: Thanks! I squashed your suggestion to the commit.

debian/gen-flat-repo-fixtures.sh
```diff
@@ -0,0 +1,27 @@
...
+cd "$OUTPUTDIR"
+(
+  for CTL in "${SRCDIR}"/asgard/*.ctl
+  do
+    equivs-build --arch ppc64 "${CTL}"
+  done
+)
+
+cd ..
+dpkg-scanpackages debian-flat/ | tee debian-flat/Packages | gzip > debian-flat/Packages.gz
```
quba42, line 19: Right now we are building a flat repo with the `fixtures` folder as its repository root, plus a distribution of `debian-flat/`.
I think we could shift things around ever so slightly to get two fixtures for the price of one:
Suppose we made `fixtures/debian-flat/` the repository root, and prepared two sets of metadata files. One in the repository root for a distribution of `/` (I think this is the most common case of flat repos in the wild), and another one in some twice nested subfolder (e.g.: `fixtrues/debian-flat/nest/fjalar/`) for a distribution of `nest/fjalar/`. (Twice nested because problems with the code often only start appearing at twice nested and we are trying to test the code).

quba42 (changes requested):
If my last comment is unclear, we can do a joint session.

quba42:
One final suggestion: For the commit message I would just use `Add debian flat repository format fixtures`. We are not actually generating any tests here (just fixtrues). And "Add" instead of "Added" because we prefer the imperative mood for commit messages. :wink: ("Added" would be preferred for release notes.)

Makefile
```diff
@@ -252,6 +259,9 @@ fixtures/debian-complex-dists: fixtures gnupghome
 fixtures/debian-missing-architecture: fixtures gnupghome
 	GNUPGHOME=$$(realpath -e gnupghome) debian/gen-missing-architecture-fixtures.sh $@
 
+fixtures/debian-flat: fixtures gnupghome
```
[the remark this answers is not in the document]
mdellweg: Actually it may be a good test, since it's important to know that the signature is found in the correct location, and this repo format messes around with all the paths.
Also thanks to Make gnupghome is only set up once per stage container anyway.

debian/gen-flat-repo-fixtures.sh
```diff
@@ -0,0 +1,27 @@
...
+
+mkdir "${OUTPUTDIR}"
+
+cd "$OUTPUTDIR"
+(
+  for CTL in "${SRCDIR}"/asgard/*.ctl
+  do
+    equivs-build --arch ppc64 "${CTL}"
+  done
+)
```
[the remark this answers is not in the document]
mdellweg: Is this whole script not run in a tmpdir environment?

debian/gen-flat-repo-fixtures.sh
ThikaXer: Done. All `*.deb` files and `Release` and `Packages` are created in `fixtures/debian-flat` and in `fixtures-debian-flat/nest/fjalar`.

Makefile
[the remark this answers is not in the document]
ThikaXer: Done

debian/gen-flat-repo-fixtures.sh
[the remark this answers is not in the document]
ThikaXer: Now (with the last push) it is run in a tmpdir. However, the line `rm *.buildinfo *.changes` still gives me the following error in the build:

In ./debian/gen-flat-repo-fixtures.sh line 18:
rm *.buildinfo *.changes
   ^-- SC2035: Use ./*glob* or -- *glob* so names with dashes won't become options.
               ^-- SC2035: Use ./*glob* or -- *glob* so names with dashes won't become options.

make: *** [lint-shellcheck] Error 123
Makefile:166: recipe for target 'lint-shellcheck' failed
Error: Process completed with exit code 2.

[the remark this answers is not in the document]
mdellweg: the tmpdir should be enough cleanup.
Also everything we don't explicitly copy to the final image will cease to exist with this ephemeral build container anyway.

IMHO no need to do extra cleanup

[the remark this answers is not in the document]
ThikaXer: Ok, done

debian/gen-flat-repo-fixtures.sh
```diff
@@ -0,0 +1,30 @@
...
+echo 'Suite: mythology
+Codename: ragnarok
+Architectures: ppc64
+Components: asgard' > Release
+
+apt-ftparchive release . >> Release
+
+mkdir -p "${OUTPUTDIR}"/nest/fjalar/
+cp -r --no-preserve=mode --reflink=auto "${TMPDIR}"/. -t "${OUTPUTDIR}"
+cp -r --no-preserve=mode --reflink=auto "${TMPDIR}"/. -t "${OUTPUTDIR}"/nest/fjalar/
```
mdellweg, line 30: @quba42 is this what you meant?
I understood, that you wanted the packages in the same directoy, but the `Packages` and `Release` files in two levels down. (i don't see a new testcase otherwise)

ThikaXer: You're right of course. Thanks. Fixed

mdellweg: Oh, i think you need to regenerate the metadata to include the correct relative paths to the packages.
Again. I don't fully understand this repository format. So it would be good, if you could at the very least say that you were able to use apt to retrieve a package from that container. (You know how to start it locally, right?)

quba42: @mdwllweg trying to locally consume the repo using APT as a test of the repo is a good idea.

On your other point, I am fairly certain it is enough to simply copy `Packages*` and `Release` to the secondary location.
The paths from `Release` to `Packages` are relative to the `Release` file, and so do not change if both are moved.
The paths within `Packages` are relative to the repository root, and so do not change.

quba42: I just tested the repo locally, and got as far as:
```
package architecture (ppc64) does not match system (amd64)
```
during package install, so I would say the repo principally works for APT.
I tested this via both sets of metadata separately.

quba42 (approved):

PR #201, opened 2020-10-30
Adding fixtures for missing architectures issue #7467

ThikaXer:
ref #7467

### DIFF
--- a/debian/gen-missing-architecture-fixtures.sh
+++ b/debian/gen-missing-architecture-fixtures.sh
@@ -1,0 +1,59 @@
+#!/bin/sh
+
+set -eux
+
+SRCDIR=$(readlink -f "$(dirname "$0")")
+TMPDIR=$(mktemp -d)
+OUTPUTDIR=$(realpath "${SRCDIR}/../$1")
+
+trap 'rm -rf "${TMPDIR}"' EXIT
+
+cd "${TMPDIR}"
+
+mkdir asgard
+(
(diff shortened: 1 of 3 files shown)


debian/gen-missing-architecture-fixtures.sh
```diff
@@ -0,0 +1,59 @@
...
+
+cp -a "${SRCDIR}/conf" .
+reprepro -C asgard includeudeb ragnarok asgard_udebs/*.udeb
+reprepro -C asgard includedeb ragnarok asgard/*.deb
+reprepro -C asgard includedeb ragnarok jotunheimr/*.deb
+reprepro -C jotunheimr includedeb ragnarok jotunheimr/*.deb
+
+# No Packages files at all for architecture armeb
+rm -r dists/ragnarok/asgard/binary-armeb
+rm -r dists/ragnarok/jotunheimr/binary-armeb
```
mdellweg, line 55: This is what makes this repository special?

What is the expected behaviour here? Failing, because the files are really expected?

mdellweg (approved):

debian/gen-missing-architecture-fixtures.sh
ThikaXer: Yes, pulp would throw an Error for architectures with missing Packages file, please also see https://github.com/pulp/pulp_deb/pull/226/ for reference.

PR #203, opened 2020-11-16
Fix workflow configuration

mdellweg:
[noissue]


daviddavis (approved):

daviddavis:
FYI, Travis had a way to [configure hostnames](https://docs.travis-ci.com/user/hosts/). I don't see the same feature for Github Actions. We could probably use localhost or maybe update the hosts file.

PR #206, opened 2020-12-21
Update fedora registry

daviddavis:


Containerfile
```diff
@@ -1,5 +1,5 @@
 # === Build fixtures (Fedora) =================================================
-FROM fedora:30 AS fedora-build
+FROM fedora:latest AS fedora-build
```
fao89, line 2: we had some issues on pulp_installer, we had to specify the registry:
https://github.com/pulp/pulp_installer/blob/master/molecule/scenario_resources/Dockerfile.j2#L9

daviddavis:
I filed an issue: https://pulp.plan.io/issues/8016

daviddavis:
I've just updated the registry to use the fedora registry. It doesn't have limits (unlike docker hub) and it does have an F30 image (unlike quay).

mdellweg (approved):

PR #207, opened 2021-01-14
Adding a fixtures.pulpproject.org tag

daviddavis:
The problem is that osci can't deploy our container due to the entrypoint. This change publishes a container image with the url hardcoded.


daviddavis:
Closing in favor of https://github.com/pulp/pulp-fixtures/pull/208

PR #208, opened 2021-01-19
Prepare fixtures container for Openshift

mscherer:
So this commit prepare for a migration of fixtures to openshift. 

I just need to take a look at the https deployment, there is a operator for that. And do the job.

### DIFF
--- a/.ci/build_image.sh
+++ b/.ci/build_image.sh
@@ -9,7 +9,7 @@ make lint
 docker build -f Containerfile -t pulp/pulp-fixtures .
 
 # Test
-docker run --rm -d -e BASE_URL=http://localhost:8000 -p 8000:80 --name pulp-fixtures pulp/pulp-fixtures
+docker run --rm -d -e BASE_URL=http://localhost:8000 -p 8000:8080 --name pulp-fixtures pulp/pulp-fixtures
 sleep 2 # it can take a couple seconds for sed to run and nginx to boot
 curl --fail localhost:8000/file/PULP_MANIFEST
 curl --fail localhost:8000/debian/dists/ragnarok/Release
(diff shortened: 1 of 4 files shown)


daviddavis:
These code changes look good to me. We link commits to our issue tracker but you can get around this by adding `[noissue]` to your commits. Also, it looks like the tests are failing.

mscherer:
```
 Step 13/26 : FROM registry.redhat.io/ubi8/ubi AS server
Get https://registry.redhat.io/v2/ubi8/ubi/manifests/latest: unauthorized: Please login to the Red Hat Registry using your Customer Portal credentials. Further instructions can be found here: https://access.redhat.com/RegistryAuthentication
```

Curious, it work on openshift. That's a bit annoying, I will need to find a way.

mscherer:
I am doing some test by using a Fedora base image, and chasing down the UBI issue (because UBI usage is on the plate of my boss). I have read the doc, it doesn't say you need a subscription: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/building_running_and_managing_containers/index#using_standard_red_hat_base_images 

And running:
```
podman run registry.access.redhat.com/ubi8/ubi cat /etc/os-release
```

seems to work fine on my Fedora 33 Silverblue laptop, and I never entered any credentials.

I guess the difference is maybe podman vs docker (and how the registry is accessed), and the credentials message is misleading.

mscherer:
Ok so the problem with the container was registry.redhat.io vs registry.access.redhat.com. I guess I used a old doc somewhere.

daviddavis:
Ah, I see the answer in one of your commits. Thanks! 

This is still a draft. Feel free to unset that when you think it's ready and I can merge.

mscherer:
So, i was waiting on more info from openshift PM, but got no answer, so I will proceed with my plan for https. I guess I shouldn't have blocked the merge on that :/

daviddavis (approved):
Looks great. Thanks!

PR #211, opened 2021-02-25
Send nginx logs on stdout/stderr

mscherer:
Since there is nothing, nginx do not send anything
anywhere, and so nothing appear in openshift.

[noissue]

### DIFF
--- a/common/nginx.conf
+++ b/common/nginx.conf
@@ -5,6 +5,8 @@ worker_processes  1;
 
 pid        /tmp/nginx.pid;
 
+error_log stderr;
+
 events {
     worker_connections  1024;
 }
@@ -30,6 +32,8 @@ http {
         listen       8080;
         server_name  localhost;
         absolute_redirect off;
+
+        access_log /dev/stdout;
 
         # if auth token after URL is '?secret' return
         # address without this token (only for urls with 'rpm')


mscherer:
While opening the PR, I realize that I am unsure of the effect on people using the container directly, maybe we should ask them.

mdellweg:
I am absolutely :+1: on this. If you run the container in the foreground while working on tests, it is very helpful to see the accesslog scrolling through a second terminal.

daviddavis:
This change looks good to me but I am not sure why it's causing our tests to fail?

mdellweg:
Can you remove the `--rm` flag in this line? Maybe this helps us to retain the logs.
https://github.com/pulp/pulp-fixtures/blob/master/.ci/build_image.sh#L12

mscherer:
Ok seems I need a review to trigger the CI

daviddavis (approved):

daviddavis:
This is strange. I've never seen the CI not automatically trigger when there's a new commit.

daviddavis:
Can you try amending your last commit so it gets a new SHA?

mdellweg:
I would not say, that we don't want to merge the "--rm" change. I think this is a good habit to keep your local box clean from stale containers, but in the CI the whole thing gets cleaned out anyway, and if we get the logs that way, we should keep it.

mscherer:
```
nginx: [emerg] "access_log" directive is not allowed here in /etc/nginx/nginx.conf:9
```

Ok, guess I am going back to read nginx doc :)

mscherer:
I removed the --rm, because the error message was captured, but hidden with "display log on error", who is collapsed by default. That's nice to show the error log in a separate section. Less nice to place it after the regular log :)

mdellweg (approved):

PR #213, opened 2021-05-10
Added a repo tha uses static_context=true for the kangaroo modules.

ggainey:
Added command-line switches to gen-patched-fixtures to support
patching of modules.yaml as well as updatinfo.

This is in support of https://pulp.plan.io/issues/8638.

[noissue]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,6 +73,8 @@ help:
 	@echo "        Create an RPM repository with modules"
 	@echo "    fixtures/rpm-with-modules-modified"
 	@echo "        Create an RPM repository with modules but without some dependencies."
+	@echo "    fixtures/rpm-modules-static-context"
+	@echo "        Create RPM Modules fixture data with static_context entries."
 	@echo "    fixtures/rpm-incomplete-filelists"
 	@echo "        Create an RPM repository with an incomplete filelists.xml."
 	@echo "    fixtures/rpm-incomplete-other"
@@ -206,6 +208,7 @@ all-fedora: \
 	fixtures/rpm-alt-layout \
 	fixtures/rpm-with-modules \
 	fixtures/rpm-with-modules-modified \
+	fixtures/rpm-modules-static-context \
 	fixtures/rpm-incomplete-filelists \
 	fixtures/rpm-incomplete-other \
 	fixtures/rpm-invalid-rpm \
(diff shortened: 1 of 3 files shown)


daviddavis (approved):
LGTM. Anyone from @pulp/rpm want to review before I merge?

Makefile
```diff
@@ -453,35 +456,38 @@ fixtures/rpm-unsigned-modified: fixtures
...
-	rpm/gen-patched-fixtures.sh $@ rpm/modules-updateinfo.patch
+	rpm/gen-patched-fixtures.sh -d $@ -f rpm/modules-updateinfo.patch
 	rm $@/kangaroo-0.3-1.noarch.rpm
 	rm $@/shark-0.1-1.noarch.rpm
 	rm $@/stork-0.12-2.noarch.rpm
 	createrepo --update $@
 	modifyrepo --mdtype=modules rpm/assets/modules.yaml "$@/repodata/"
 
+fixtures/rpm-modules-static-context: fixtures
+	rpm/gen-patched-fixtures.sh -d $@ -f rpm/modules-static-context.patch -t module
```
goosemania, line 489: My understanding is that it will result in the modules.yaml with mixed format, aka some modules have static_context, some do not. Correct?  It's good, quite a realistic situation when folks start mixing up content from different repos.

rpm/modules-static-context.patch
```diff
@@ -0,0 +1,30 @@
...
+   summary: Kangaroo 0.2 module
+   description: >-
+@@ -51,6 +53,7 @@
+   name: duck
+   stream: 0
+   version: 201809302113907
++  static_context: false
+   context: deadbeef
+   arch: noarch
+   summary: Duck 0.8 module
```
goosemania, line 30: The modules to set static_context for, were they chosen with the dep solving testing in mind?
It's a pain to change fixtures, because it often will require adjusting tests which use it.
I think we either need to think what would be the best set of modules to have static_context set to test dep solving properly when we add this feature, or we will need a new fixture for dep solving when we get to it. And it's ok, just saying it out loud to be sure we are on the same page.

Makefile
ggainey: Yes - that's deliiberate on my part. I don't see anything in the spec that forbids it, so I figure it'll happen eventually!

goosemania (approved):

PR #214, opened 2021-06-18
Add different fields in deb and Package index

m-bucher:
Preparation for unit-tests in https://github.com/pulp/pulp_deb/pull/294
Originally I wanted to remove `Description` from the Package-Index, because it should not be necessary if `Description-md5` is present.
But pulp_deb does not see it this way right now :stuck_out_tongue_winking_eye: 
Also we have to add Translation files to the Metadata if I understood correctly: https://wiki.debian.org/DebianRepository/Format#Description

### DIFF
--- a/debian/conf/override.ragnarok
+++ b/debian/conf/override.ragnarok
@@ -1,0 +1,6 @@
+* Task Æsir
+odin Task Æsir
+odin Description-md5 654ca349ba8dade6580ba77c71d67eff
+thor Task Æsir
+thor Bugs https://bilskirnir.example.com/bugs
+thor Description-md5 f5d3997e79acfeeb7a6ff4aa71ab65ac
--- a/debian/conf/distributions
+++ b/debian/conf/distributions
@@ -3,12 +3,14 @@ suite: mythology
 Architectures: armeb ppc64
 Components: asgard jotunheimr
 UdebComponents: asgard
+DebOverride: override.ragnarok
 SignWith: 6EDF301256480B9B801EBA3D05A5E6DA269D9D98
 
 codename: nosuite
 Description: This is a valid Debian repository.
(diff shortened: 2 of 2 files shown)


quba42:
Right now syncing translation files is completely disabled in pulp_deb. But yes, if we want to fix the code and re-enable this feature we need some fixtures to test it.

debian/conf/override.ragnarok
```diff
@@ -0,0 +1,6 @@
+* Task Æsir
+odin Task Æsir
+odin Description-md5 654ca349ba8dade6580ba77c71d67eff
+thor Task Æsir
+thor Bugs https://bilskirnir.example.com/bugs
+odin Description-md5 654ca349ba8dade6580ba77c71d67eff
```
mdellweg, line 6: should this be thor?

m-bucher: Oh, nice catch, of course  :facepalm:

mdellweg: And still i have no idea what it's for...

dralley:
What is the status of this PR?

quba42:
> What is the status of this PR?

It was waiting for the corresponding pulp_deb PR, which was waiting for another PR. That other PR is now merged, things are unblocked and we should get this done soon.

quba42:
Code changes are OK, needs to be tested one more time against the current pulp_deb main branch?

m-bucher:
I ran tests against main-branch with locally created pulp-fixtures and apart from some unittest-fails and an expected signing-service-test fail it ran through.
So from my side, this is good to go.

quba42 (approved):
I am fine with the code changes, and the reported manual tests by @m-bucher, so I will go ahead and merge this.

PR #215, opened 2021-07-19
Add new repo for packages that exercise more parts of the RPM metadata

dralley:
Note that this specfile and the scripts are from a personal project done outside of work hours, so I retain the right to use it under a different license (probably MPLv2, maybe MIT + Apache)

[noissue]


rpm-new/build_repos.sh
```diff
@@ -0,0 +1,30 @@
...
+output_dir="$(realpath "$1")"
+
+# Complex repository - use all metadata features
+
+mkdir -p "$output_dir"
+
+while IFS= read -r filename
+do
+  cp --no-preserve=mode --reflink=auto "$assets_dir/packages/$filename" "$output_dir/$filename" ;
+done < "$assets_dir/complex_repo_pkglist.txt"
```
dralley, line 15: We should make this more generic if / when we re-use it but for now this is fine.

dkliban (approved):

PR #219, opened 2021-07-23
Update complex-pkg fixture

dralley:
[noissue]


rpm-new/build_repos.sh
```diff
@@ -26,5 +23,10 @@ createrepo_c \
   --retain-old-md=0 \
   --simple-md-filenames \
   --no-database \
-  --pkglist="$assets_dir/complex_repo_pkglist.txt" \
-  "$output_dir"
+  "$packages_dir"
+  # --pkglist="$assets_dir/complex_repo_pkglist.txt" \
```
ggainey, line 27: Why leave this in as a comment?

dralley: For some reason it's not working, but the repo creation works fine, and at this point I want to fix pulp_rpm first and fix the fixture later

ggainey: Ah! I get it, kk. When you have a sec, open an issue for the 'fix it later', if you would.

ggainey (approved):

goosemania (approved):

PR #223, opened 2021-09-20
Add RPM metadata only fixture

pavelpicka:
for ACS test

closes: #9422
https://pulp.plan.io/issues/9422

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -125,6 +125,8 @@ help:
 	@echo "        Create RPM fixture data with signed packages."
 	@echo "    fixtures/rpm-unsigned"
 	@echo "        Create RPM fixture data with unsigned packages."
+	@echo "    fixtures/rpm-unsigned-meta-only"
+	@echo "         Creates RPM fixture with metadata only."
 	@echo "    fixtures/rpm-unsigned-modified"
 	@echo "        Create RPM fixture without the whale package."
 	@echo "    fixtures/rpm-updated-updateinfo"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -490,6 +493,10 @@ fixtures/rpm-signed: fixtures gnupghome
 fixtures/rpm-unsigned: fixtures
 	rpm/gen-fixtures.sh $@ rpm/assets
 
+fixtures/rpm-unsigned-meta-only: fixtures fixtures/rpm-unsigned
+	mkdir -p $@
+	cp -r fixtures/rpm-unsigned/repodata $@/
```
daviddavis, line 498: I believe the `-r` option for cp is discouraged.

```suggestion
	cp -R fixtures/rpm-unsigned/repodata $@/
```

pavelpicka: any mentioned in ```man``` but for consistency of ```Makefile``` looks much better.

daviddavis: The behavior for `-r` is not defined by POSIX and can vary from system to system. On some systems, it's equivalent to `-R` but on other systems (e.g. FreeBSD) it's equivalent to `-RL`.

daviddavis (approved):

Makefile
pavelpicka: thank you

ipanova:
looks good, can you add it to the fedora-all target as well, like https://github.com/pulp/pulp-fixtures/commit/f344457f949b0426524b29d0f5e8dc65a23c15b8

ipanova:
disregard, i'm blind :)

ipanova (approved):

PR #229, opened 2022-06-21
Add obsolete to modulemd

pavelpicka:
add obsolete type to modulemd yaml file

[noissue]


pavelpicka:
another fixture will be created instead of this one

PR #233, opened 2022-11-25
Add Debian 11 compatible test fixtures

hstct:
This PR tries to solve two issues the current debian fixtures have:

- There are not Packages set to `Architecture: all`.
- There is no repository with the [format](https://wiki.debian.org/DebianRepository/Format#No-Support-for-Architecture-all) `No-Support-for-Architecture: all`

Both are special cases in `pulp_deb` and they should be tested.
 
[noissue]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,8 @@ help:
 	@echo "        Create Debian repository fixtures with missing Packages files for some architectures."
 	@echo "    fixtures/debian-flat"
 	@echo "        Create Debian repository fixtures in flat repository format."
+	@echo "    fixtures/debian-mixed"
+	@echo "        Create Debian 11 compatible repository fixtures."
 	@echo "    fixtures/docker"
 	@echo "        Create Docker fixture data."
 	@echo "    fixtures/drpm-signed"
@@ -204,6 +206,7 @@ all-debian: \
 	fixtures/debian-complex-dists \
 	fixtures/debian-missing-architecture \
 	fixtures/debian-flat \
+	fixtures/debian-mixed \
 
 all-fedora: \
 	fixtures/diff-name-same-content \
@@ -308,6 +311,9 @@ fixtures/debian-missing-architecture: fixtures gnupghome
 
 fixtures/debian-flat: fixtures
 	debian/gen-flat-repo-fixtures.sh $@
+
+fixtures/debian-mixed: fixtures gnupghome
+	GNUPGHOME=$$(realpath -e gnupghome) debian/gen-mixed-fixtures.sh $@
 
 fixtures/docker: fixtures
 	docker/gen-fixtures.sh $@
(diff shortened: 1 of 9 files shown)


hstct:
The script is a bit hacky mainly due to the fact that I'm not able to cleanly produce a repository with `binary-all` components using `reprepro`. If anyone knows a better way I would love to know.

Makefile
```diff
@@ -32,6 +32,8 @@ help:
 	@echo "        Create Debian repository fixtures with missing Packages files for some architectures."
 	@echo "    fixtures/debian-flat"
 	@echo "        Create Debian repository fixtures in flat repository format."
+	@echo "    fixtures/debian11"
```
quba42, line 35: Should we perhaps call this `debian-mixed` instead of debian11, since we refer to this repo format as "mixed" from within the pulp_deb code? In principle this format could be used by repos other than debian11 (even though debian11 is the main example I am aware of).

debian/conf/distributions
```diff
@@ -18,3 +18,8 @@ Description: This repository is valid, but contains no packages!
 Architectures: armeb ppc64
 Components: asgard jotunheimr
 SignWith: 6EDF301256480B9B801EBA3D05A5E6DA269D9D98
+
+codename: muspelheim
+Architectures: amd64 armeb ppc64
```
quba42, line 23: Do we need two architectures besides `armeb/all` here, or does this needlessly complicate this fixture that is not about testing lots of different architectures, but about testing the `all` architecture in particular?

debian/gen_complex_dists_fixtures.sh
```diff
@@ -14,8 +14,7 @@ mkdir asgard
 (
   cd asgard
 
-  for CTL in "${SRCDIR}"/asgard/*.ctl
-  do
+  for CTL in "${SRCDIR}"/asgard/*.ctl; do
```
quba42, line 17: Do we want unrelated bash style changes as part of this commit?

quba42 (changes requested):
I was able to test these fixtures by building them locally, adding the repo to my laptops `/etc/apt/sources.list` and installing a package from them. <= Looks like they all work!

The built fixtures also appear to satisfy the spec for `No-Support-For-Architecture-All: Packages` style repos (as far as I can tell).

The way we work around the limitations of reprepro in order to build this repo structure is acceptable to me. (I am not aware of any better tools for the job).

We could consider also adding architecture all packages to some of the other existing fixtures (but that would break existing tests, so it would probably be better to do that on a separate PR if at all).

I have some small nitpicks around how we might adjust the naming, but I can be convinced otherwise. Once we reach a consensus on those I think this is ready to go!

quba42:
> Also no idea about the norse mythology names I just stuck with the theme and googled something appropriate hopefully

@mdellweg will be pleased!

quba42 (approved):
Re-tested building these locally, and compared with the previous iteration for correctness.

LGTM!

quba42:
Since this adds an entirely new fixture, I don't see how this could harm anything existing, so I will now merge.

PR #240, opened 2023-06-19
Fix typo in package control file

hstct:
[noissue]


mdellweg (approved):

quba42:
We are 90% sure this coupled with the recent re-build using `debian:bullseye` instead of `debian:stretch` is responsible for e.g.: the following test failure on pulp_deb: https://github.com/pulp/pulp_deb/actions/runs/5301400952/jobs/9595659720?pr=798

quba42 (approved):

PR #241, opened 2023-09-28
Add source package files to the main debian repository

hstct:
[noissue]


hstct:
Since we are still occasionally using the hosted fixtures it would be good to have it here as well. It's also a good collection of scripts to build the test data :)

dralley (approved):

PR #243, opened 2023-10-28
Implement PyPI version json API & added two packages for dependency syncing

gerrod3:
[noissue]

### DIFF
--- a/python/pypi-assets/urlformatter.py
+++ b/python/pypi-assets/urlformatter.py
@@ -40,7 +40,9 @@ def main():
 def format_url(metadata, base_url):
     """Format the distribution urls in metadata to point to the distributions
     hosted on base_url"""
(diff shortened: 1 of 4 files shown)


python/pypi-assets/urlformatter.py
```diff
@@ -40,7 +40,9 @@ def main():
 def format_url(metadata, base_url):
     """Format the distribution urls in metadata to point to the distributions
     hosted on base_url"""
-    for distributions in chain(metadata["releases"].values(), [metadata["urls"]]):
+    release_urls = metadata.get("releases", {}).values()
+    url_urls = [metadata.get("urls", [])]
```
mdellweg, line 44: Assuming `metadata["urls"]` is already a list (as the default suggests):
```suggestion
    url_urls = metadata.get("urls", [])
```

mdellweg (approved):

Issue #242, opened 2023-10-09
zstd compressed RPM fixtures


dralley:
Will require createrepo_c 1.0 to be present, so it'll likely need to wait until Fedora 39 is released (which should be soon now)

PR #244, opened 2023-11-16
Update fixture generation and add a zstd metadata fixture

dralley:
Closes https://github.com/pulp/pulp-fixtures/issues/245
Closes https://github.com/pulp/pulp-fixtures/issues/242


dralley:
@quba42 Any idea why the Debian portion is failing to build?  I assume this is the line: https://github.com/pulp/pulp-fixtures/actions/runs/6885272791/job/18729164889?pr=244#step:5:1941

I'm not sure if this has to do with swapping out the previous key though, because as far as I can tell I fixed it up in all the correct places?

.github/workflows/issue_checker.yaml
```diff
@@ -5,13 +5,11 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
```
mdellweg, line 8: :+1:

common/GPG-KEY-pulp-qe
mdellweg, line 1: What called for this change?
I believe some CI is using the key directly from here. So changing it's name is probably a bad idea.
As is changing the key user-id.

Containerfile
```diff
@@ -2,18 +2,21 @@
...
               ostree \
               patch \
+              python3-pip \
               python3-jinja2-cli \
               rpm-build \
               rpm-sign \
               rsync
 
+# the default createrepo_c provided by Fedora has legacy hashes disabled, the one
+# on PyPI does not (because we need it)
```
mdellweg, line 18: What happens if that package would be upgraded also? Want to add a version constraint?

dralley: Nah.  It's actually good for something like that to make a bit of noise early on.

But also, the Python package is basically our responsibility in the first place, so it's unlikely to change out from under us.

common/GPG-KEY-pulp-qe
dralley: The key is ancient and uses algorithms that recent OS' consider legacy, and refuse to do anything with.

dralley: https://github.com/pulp/pulp-fixtures/issues/245

mdellweg (approved):

dkliban (approved):

PR #246, opened 2023-11-17
Delete old fixture signing keys

dralley:
[noissue]


quba42:
I am pretty sure these are used/needed for pulp_deb tests. Please give me some time to investigate.

hstct:
We actually don't it since we are using the pulpcore fixtures for this and only need to update the key name.

Once it gets released by pulpcore the tests should work here https://github.com/pulp/pulp_deb/pull/956

mdellweg:
I still think it won't hurt to keep them around for a month or so.

mdellweg:
I fail to see the pressure. But maybe we only see what we missed when actually remove them.

mdellweg (approved):

PR #247, opened 2023-11-21
Restore DRPM fixtures

dralley:
Even though we don't support it (and don't plan to), I forgot that we do still have a test with an expectation that the repository is rejected in certain modes.  We could probably delete the test, but I'd prefer to just restore the fixture, it's not so big a deal.


dralley:
Nope, because the Python createrepo_c package doesn't have deltaRPM support.  This is no longer worth it.

PR #248, opened 2023-11-21
Don't automatically compress metadata files via modifyrepo_c

dralley:
[noissue]


dralley:
I believe the tests on https://github.com/pulp/pulp_rpm/pull/3323 are broken because of this

PR #253, opened 2024-08-07
Add empty debian repo

dosas:


debian/gen-empty-fixtures.sh
```diff
@@ -0,0 +1,21 @@
...
+
+echo 'Suite: mythology
+Codename: ragnarok
+Architectures: ppc64
+Components: asgard' > Release
+
+apt-ftparchive release . >> Release
+
+mkdir -p "${OUTPUTDIR}"/dists/fjalar/
+cp -r --no-preserve=mode --reflink=auto "${TMPDIR}"/Release -t "${OUTPUTDIR}"/dists/fjalar/
```
mdellweg, line 21: This looks like a hell of a command to copy a single file. Am I missing something?
Also what about the output if apt-ftparchive here?
If the "empty repository" is just a release file, we can short cut this a lot i think.
Also should the name of the dists directory not match the codename?

hstct: the cp command is just copy paste from the other scripts so for consistency sake

it's an empty repository without any packages iirc but it should include a release file

PR #255, opened 2025-06-20
Fix CI

jobselko:
This PR fixes the broken CI build by addressing a ShellCheck warning and pinning Fedora to version 40 and createrepo_c to version 1.2.0, instead of using their latest versions.

A proper fix that enables support for the latest versions is tracked in #256.


gerrod3 (approved):
Nice! Thanks for sticking with it.

PR #260, opened 2026-05-01
Add RPM packages for testing unsigned, signed, & multiple-signed

dralley:
Including with a PQC-signed RPM

[noissue]

Assisted-By: claude-opus-4.6


rpm-new/assets/specs/build_and_sign.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
```
dralley, line 1: This could be much simpler if I could just use my own tools, however, unfortunately I figure for something like this it's best that we actually use the "standard" tools despite how much more convoluted it makes things.

dralley:
And also: statically commit the md5 and sha1 repos since they're hard to make unless you build createrepo_c from source with the `LEGACY_HASHES` option

mdellweg (approved):

